PR #3, opened 2022-09-21
Vision conversion

bramstoeller:


pyproject.toml
```diff
@@ -76,11 +77,11 @@ power_grid_model_io = ["*.yaml"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests/unit"]
-addopts = ["--cov=src/power_grid_model_io", "--cov-report", "term", "--cov-fail-under=20"]
+addopts = ["--cov=src/power_grid_model_io", "--cov-report", "term", "--cov-report", "html:python_cov_html", "--cov-report", "xml:python_coverage.xml", "--cov-fail-under=20"]
```
bramstoeller, line 80: Do we really need the html covarage report?
Can we name the coverage report `coverage.xml`; this is a python only project.

bramstoeller (commented):
In the .gitignore file: I think it's a bit messy to just ignore everything that could come up. Can we be a bit more selective?

pyproject.toml
```diff
@@ -76,11 +77,11 @@ power_grid_model_io = ["*.yaml"]
...
 
 [tool.black]
 line-length = 120
-target-version = ["py39", "py310", "py311"]
+target-version = ["py39"]
```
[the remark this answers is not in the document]
TonyXiang8787: Done

pyproject.toml
TonyXiang8787: xml is needed later for sonarcloud.

html is not per se needed. I personally find it nice to read the coverage report in html.

TonyXiang8787: removed python prefix.

TonyXiang8787:
@bramstoeller  two general comments:

1. Did you check licenses of all the dependencies. Are they compatible with our MPL2 license?
2. I am not so sure about optional dependencies. Can we just include all the dependencies as default?

pyproject.toml
```diff
@@ -40,7 +43,27 @@ dev = [
...
+    "openpyxl",
+    "pyyaml",
+    "typer[all]",
+]
+
+[project.urls]
+Home-page = "https://github.com/alliander-opensource/power-grid-model-io"
+
+[project.scripts]
+convert = "power_grid_model_io.cli.convert:app"
```
TonyXiang8787, line 65: the script command will be inject into the environment. I would call them with a prefix like `pgm_convert` or `pgm_valdiate`. Otherwise we may have a name conflict with other system command.

src/power_grid_model_io/utils/filters.py
```diff
@@ -0,0 +1,151 @@
...
+    "YN": WindingType.wye_n,
+    "D": WindingType.delta,
+    "Z": WindingType.wye,
+    "ZN": WindingType.wye_n,
+}
+
+T = TypeVar("T")
+
+
+def relative_no_load_current(i: float, p: float, s_nom: float, u_nom: float) -> float:
```
TonyXiang8787, line 31: @bramstoeller  since the conversion performance will be essential (for pandapower). We need to have vectorized version of all these filter functions. The input/output should be numpy arrays.

src/power_grid_model_io/data_stores/base_data_store.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        self._log = structlog.get_logger(type(self).__name__)
+
+    @abstractmethod  # pragma: no cover
+    def load(self) -> T:
+        """
+        The method that loads the data from one or more sources and returns it in the specified format.
+        Note that the load() method does not recieve a reference to the data source(s); i.e. the data souce(s)
+        should be set in the constructor, or in a separate member method.
+        """
```
TonyXiang8787, line 33: an empty method should have a `pass` in the body.

src/power_grid_model_io/cli/convert.py
```diff
@@ -0,0 +1,104 @@
...
+
+import structlog
+from power_grid_model.validation import errors_to_string, validate_input_data
+
+from power_grid_model_io.converters.gaia_excel_converter import GaiaExcelConverter
+from power_grid_model_io.converters.pgm_converter import PgmConverter
+from power_grid_model_io.converters.vision_excel_converter import VisionExcelConverter
+from power_grid_model_io.utils.modules import import_optional_module
+
+typer = import_optional_module("cli", "typer")
```
petersalemink95, line 22: I'd prefer to specify the parameters here, e.g. `typer = import_optional_module(extra="cli", module="typer")`

src/power_grid_model_io/data_stores/base_data_store.py
bramstoeller: Not necessarily; `pass` is just a placeholder to fix the syntax. In this case, the comment serves as a placeholder. I'm not entirely sure, but I guess the liner (pylint) would complain about an extra `pass`.

src/power_grid_model_io/cli/convert.py
bramstoeller: Yes, I agree. I'll fix it.

pyproject.toml
bramstoeller: Yes, fair point. I'll fix it.

src/power_grid_model_io/utils/filters.py
bramstoeller: I agree, partially.

My rationale was: Create a way to define functions in an understandable way. These mapping functions are much easier to read and understand than vectorized function. Currently we didn't have any efficiency constraints, but indeed, they'll come up in the near future. Let's create new stories for it.

Peter and I do have some ideas about efficiency (for pandapower), let's discuss it in a refinement meeting.

bramstoeller:
> @bramstoeller two general comments:
> 
> 1. Did you check licenses of all the dependencies. Are they compatible with our MPL2 license?
No, I didn't. I'll check them.
> 2. I am not so sure about optional dependencies. Can we just include all the dependencies as default?
It was a requirement from one of our users to "please make sure that the conversions library doesn't get super heavy".
We could discuss how to organize the extra requirements.

TonyXiang8787:
@bramstoeller 
https://github.com/Alliander/python_scientific_computing/blob/master/Tabular_data_manipulation_in_python.ipynb

examples/pgm_conversion.py
```diff
@@ -0,0 +1,29 @@
...
+from power_grid_model import PowerGridModel
+
+from power_grid_model_io.converters.pgm_converter import PgmConverter
+
+DATA_DIR = Path(__file__).parent / "data"
+
+if __name__ == "__main__":
+    structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(logging.DEBUG))
+
+    input_file = DATA_DIR / "1os2msr" / "input.json"
```
petersalemink95, line 19: I think we should rename these examples

examples/pgm_conversion.py
```diff
@@ -0,0 +1,29 @@
...
+    output_file = DATA_DIR / "1os2msr" / "sym_output.json"
```
petersalemink95, line 20: This one as well

examples/pgm_conversion.py
petersalemink95: 1os2msr doesn't mean anything in English

.gitignore
```diff
@@ -2,17 +2,146 @@
...
 
-# Generated files
-*.egg-info
-
-# temporary_files
+# Byte-compiled / optimized / DLL files
 __pycache__/
-.mypy_cache/
-.pytest_cache/
+*.py[cod]
```
petersalemink95, line 7: @TonyXiang8787 This comment counts for all things that have changed in the `.gitignore`.

Why did you add so many things to ignore? I think this is an overkill. We should only add things to the gitignore when you plan to use is and don't want it to be commited. I think it doesn't make sense to try to add all the things that you can think of to the gitignore.

docs/converters/tabular_converter.md
```diff
@@ -0,0 +1,186 @@
+<!--
+SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model IO project <dynamic.grid.calculation@alliander.com>
+
+SPDX-License-Identifier: MPL-2.0
+-->
+# Tabular converter
+Tabular data can come from Excel files, a set of CSV files, GNF files, databases, pandas DataFrames, etc etc.
+The similarity between all tabular data is that it contains multiple `tabled`,
```
petersalemink95, line 8: *typo: tables

docs/converters/tabular_converter.md
```diff
@@ -0,0 +1,186 @@
...
+
+substitutions:
+  ".*Switch state":
+    "off": 0
+    "on": 1
+```
+
+## Grid
+For each `table`, the target PGM `component` is listed (e.g. Nodes: node, Cables: line).
+The for each PFM `column` the source column is supplied (e.g. u_rated: Unom, from_status: From.SwitchStatus).
```
petersalemink95, line 45: *typo PGM

docs/converters/tabular_converter.md
```diff
@@ -0,0 +1,186 @@
...
+but in many cases the mapping is a bit more complex.
+You can use the following `column` definitions: 
+
+  * Column name `str`
+    ```yaml
+    from_node: From.Number
+    ```
+  * First matching column name `str`
+    ```yaml
+    from_node: Inverter.Pnom | Inverter.Snom
```
petersalemink95, line 59: `from_node` should be `p_specified`

examples/pgm_conversion.py
bramstoeller: Done, I've called it tiny-net.

docs/converters/tabular_converter.md
```diff
@@ -0,0 +1,186 @@
...
+        table["Inverter.Pnom"] if "Inverter.Pnom" in table else table["Inverter.Snom"]
+      ),
+      table["Inverter.Snom"],
+      1.0
+    )
+    ```
+## Units
+Power Grid Model uses SI units (e.g. "W" for Watts),
+but source data may be supplied in different units (e.g. "MW" for Mega Watts).
+Iff units are supplied in the tabular data,
```
petersalemink95, line 107: *typo if

.gitignore
TonyXiang8787: The `.gitignore` is generated by GitHub itself. When you select a language, e.g. Python, GitHub will generate a standard `.gitignore` for common files which are ignored for Python project.

TonyXiang8787: Usually when you create a new repository from GitHub, it will generate a automatic `.gitignore` based on language selection.

bramstoeller: Not if you select .gitignore template: **None** 😛 , like I did 😉

petersalemink95: I'm not convinced of using the auto generated `.gitignore`. I think it looks messy to just have 100+ random things to ignore

TonyXiang8787: @bramstoeller @petersalemink95 it appears that both of you have similar opinion. I agree with you on this and you can delete most of the changes.

petersalemink95 (approved):

PR #13, opened 2022-09-27
Move filters to separate module

bramstoeller:


TonyXiang8787:
@bramstoeller  Vision/Gaia will have very similar, mostly exactly the same filters. I would call them `phase_to_phase` filters.

tests/unit/filters/test_phase_to_phase.py
```diff
@@ -0,0 +1,128 @@
...
+    assert actual == approx(expected) or (np.isnan(actual) and np.isnan(expected))
+
+
+@mark.parametrize(
+    ("wind_speed", "expected"),
+    [
+        (0.0, 0.0),
+        (1.5, 0.0),
+        (3.0, 0.0),  # cut-in
+        (8.5, 125000.0),
```
nitbharambe, line 54: I seem to not get this. Power is supplied in 3 to 14 for default values.
At 8.5, 
power = 1e6 * ((8.5-3)/(14-3))^3 = 324951.171875

bramstoeller: Hm... on my calculator 1e6 * ((8.5-3)/(14-3))^3 = 1e6 * (5.5/11)^3 = 1e6 * 0.5^3 = 125000.0

tests/unit/filters/test_phase_to_phase.py
```diff
@@ -0,0 +1,128 @@
...
+        ("Dy7", WindingType.delta),
+        ("Dyn8", WindingType.delta),
+        ("Dd9", WindingType.delta),
+    ],
+)
+def test_get_winding_from(code: str, winding_type: WindingType):
+    assert get_winding_from(code) == winding_type
+
+
+def test_get_winding_from__exception():
```
nitbharambe, line 85: double underscore

bramstoeller: Yes, on purpose. `exception` is a postfix for `test_get_winding_from`.
Some peaple would actually write `test__get_winding_from__exception`, but `test_get_winding_from_exception` is also perfectly valid. I don't have a strong preference...

tests/unit/filters/test_phase_to_phase.py
```diff
@@ -0,0 +1,128 @@
...
+        ("Dd9", WindingType.delta),
+    ],
+)
+def test_get_winding_to(code: str, winding_type: WindingType):
+    assert get_winding_to(code) == winding_type
+
+
+def test_get_winding_to__exception():
+    with raises(ValueError):
+        get_winding_to("XNd11")
```
nitbharambe, line 110: Also this should check to winding right?

PR #17, opened 2022-10-06
Unit tests for converters

petersalemink95:


bramstoeller (dismissed):

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -179,8 +216,22 @@ def _convert_col_def_to_attribute(
         component: str,
         attr: str,
         col_def: Any,
-        extra_info: Optional[ExtraInfoLookup],
+        extra_info: Optional[ExtraInfoLookup] = None,
```
bramstoeller, line 219: I understand this urge, but I try not to use default values for private functions. I think it's better to be more explicit and let the caller explicitly state `extra_info=None`. Currently this function is only called from `_convert_table_to_component()`, which forwards the `extra_info` attribute it receives, that function receives it from `_parse_data()`* and that function receives it from `load_input_data()`, `load_update_data()` etc. By providing a default value, one might forget to forward the `extra_info` attribute.

In those high level functions, it is decided that `extra_info` is only loaded for input data. I'm not sure that that's correct, but because I found it too complex to handle extra_info for batch datasets, I decided to implement it only for input data for now, and we can think of more complex cases later on.

* Note that `_parse_data()` also has a default `None` value for the `extra_info` attribute, but it shouldn't.

bramstoeller: @petersalemink95, how do you think about this? In general I would like to avoid default values for attributes of private methods.

pyproject.toml
```diff
@@ -78,7 +78,7 @@ power_grid_model_io = ["config/**/*.yaml"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests/unit"]
-addopts = ["--cov=power_grid_model_io", "--cov-report=term", "--cov-report=html:cov_html", "--cov-fail-under=56.93"]
+addopts = ["--cov=power_grid_model_io", "--cov-report=term", "--cov-report=html:cov_html", "--cov-fail-under=95.76"]
```
bramstoeller, line 81: On my machine, the coverage is 82.86.

tests/unit/converters/test_pgm_json_converter.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.fixture
+def pgm_batch_data():
+    line = initialize_array("update", "line", (3, 2))
+    return {"line": line}
+
+
+@pytest.fixture
+def pgm_sparse_batch_data():
```
bramstoeller, line 55: I don't think this is a valid indptr, did you mean `[0, 1, 2, 3]` ?

tests/unit/converters/test_pgm_json_converter.py
```diff
@@ -0,0 +1,158 @@
...
+    assert [400.0, 400.0] in pgm_data["node"]["u_rated"]
+
+    # test for batch dataset
+    pgm_batch_data = converter._parse_data(data=structured_batch_data, data_type="update")
+    assert len(pgm_batch_data) == 1
+    assert (pgm_batch_data["sym_load"]["indptr"] == np.array([0, 1, 3])).all()
+    assert (pgm_batch_data["sym_load"]["data"]["id"] == [3, 3, 4]).all()
+    assert (pgm_batch_data["sym_load"]["data"]["p_specified"] == [1.0, 2.0, 3.0]).all()
+
+    # TODO include extra_info
```
bramstoeller, line 78: Did you plan to include this in another PR, or in this one?

tests/unit/converters/test_pgm_json_converter.py
```diff
@@ -0,0 +1,158 @@
...
+
+def test_converter__parse_dataset(converter: PgmJsonConverter, structured_input_data):
+    pgm_data = converter._parse_dataset(data=structured_input_data, data_type="input")
+
+    assert len(pgm_data) == 1
+    assert len(pgm_data["node"]) == 2
+    assert [1, 2] in pgm_data["node"]["id"]
+    assert [400.0, 400.0] in pgm_data["node"]["u_rated"]
+
+    # TODO include extra_info
```
bramstoeller, line 89: Did you plan to include this in another PR, or in this one?

bramstoeller (approved):
Very nice job! All seems fine to me. Some small remarks, please review those before merging.

I'll approve the PR right now, so you can merge it when you are done.

PR #42, opened 2022-10-25
Feature/csv dir store

bramstoeller:


src/power_grid_model_io/data_stores/csv_dir_store.py
```diff
@@ -0,0 +1,55 @@
...
+        self._dir_path = Path(dir_path)
+        self._csv_kwargs: Dict[str, Any] = csv_kwargs
+        self._header_rows: List[int] = [0]
+
+    def load(self) -> TabularData:
+        """
+        Create a lazy loader for all CSV files in a directory and store them in a TabularData instance.
+        """
+
+        def lazy_csv_loader(csv_path: Path) -> Callable[[], pd.DataFrame]:
```
petersalemink95, line 38: Why do we have these two local functiond here (`lazy_csv_loader`/`csv_loader`) when we could just do the call of line 40 in line 46?

src/power_grid_model_io/data_types/tabular_data.py
```diff
@@ -50,7 +52,7 @@ def get_column(self, table_name: str, column_name: str) -> pd.Series:
         """
         Select a column from a table, while applying unit conversions and value substitutions
         """
-        table_data = self._data[table_name]
+        table_data = self[table_name]
```
petersalemink95, line 55: Does this work?

src/power_grid_model_io/data_stores/csv_dir_store.py
bramstoeller: It returns the bound function `csv_loader`, which is called later.

src/power_grid_model_io/data_types/tabular_data.py
bramstoeller: Yep. It is a synonym for `self.__getitem__(table_name)`

bramstoeller: The `__getitem__()` duntion returns `self._data[table_name]`, but may call the lazy_csv_loader first:
```python
def __getitem__(self, table_name: str):
    if callable(self._data[table_name]):
        self._data[table_name] = self._data[table_name]()
    return self._data[table_name]
```

petersalemink95 (approved):

PR #48, opened 2022-10-26
Feature/pp if then

bramstoeller:


bramstoeller:
I guess we could implement this more efficiently, but it works for now...

TonyXiang8787:
@bramstoeller @nitbharambe @Laurynas-Jagutis 

My opinion on loads is to always create three loads (const power, current, impedance) to represent one load in pandapower. This should not bring much performance hit. Appliances are really cheap to run in PGM.

And then the solo purpose of if-then logic is for the switches.

PR #50, opened 2022-10-27
markdown for converters

nitbharambe:


docs/converters/converter.md
```diff
@@ -0,0 +1,50 @@
...
+-->
+
+# Converters
+
+There are 4 types of converters present as of now. PGM JSON Converter, VisonExcelConverter, GaiaExcelConverter and Pandapower Converter.
+Among them, VisonExcelConverter and GaiaExcelConverter are [tabular converters](tabular_converter.md) for excel exports of vision and gaia respectively.
+
+PGM JSON converter converts to and from the power-grid-model JSON data format. 
+This json format makes it convenient to evaluate the grid data in perspective of input to power-grid-model. 
+More details about the JSON format are mentioned in {external:py:class}`power_grid_model.utils` and example of it is in [Make Test Dataset](power-grid-model:exapmles/Make%20Test%20Dataset.ipynb) in power-grid-model repository.
```
petersalemink95, line 14: Should {external:py:class} be visible in de markdown? I guess something goes wrong here ;)

TonyXiang8787: It is needed for generation of read the docs.

petersalemink95: Hmm okay, I don't like that it is showing in the rendered markdown, but if this is the way to go with readthedocs then I think it's okay

petersalemink95:
I general I think it would be nice to have a part about the way the converters are structured. I.e. the inheritance from base tot tabular to ...... converters. In Addition, that the usable functions can be found in the BaseConverter and that the private functions `_load_data`, `_parse_data` and `_serialize_data` are the functions that are overloaded in child classes, and should also be overloaded if users want to create their own converter classes.

If you need a bit of help with this part, please ask me or Bram.

docs/converters/converter.md
nitbharambe: Yes, things like these are visible in all of the docs of both pgm and pgm-io.  No way around it. We should now always try to redirect users to the readthedocs site as there is reading via markdowns is not as fluid as it should be. Other libraries do the same as well.

petersalemink95: Thanks

petersalemink95 (approved):

PR #52, opened 2022-10-28
Various tweaks and fixes, cherry picked from an obsolete branch

bramstoeller:


src/power_grid_model_io/filters/__init__.py
```diff
@@ -58,3 +67,53 @@ def complex_inverse_imaginary_part(real: float, imag: float) -> float:
...
     """
     return (1.0 / (real + 1j * imag)).imag
+
+
+def get_winding(winding: str, neutral_grounding: bool = True) -> WindingType:
+    """
+    Return the winding type as an enum value, based on the string representation
+    """
+    winding_type = WINDING_TYPES[winding.upper()]
+    if not neutral_grounding:
```
petersalemink95, line 77: If it possible to have a `winding_type` with neutral like `wye_n` with `neutral_grounding = False?`

tests/unit/filters/test_filters.py
```diff
@@ -9,8 +10,11 @@
 from power_grid_model_io.filters import (
```
petersalemink95, line 10: There is no unit test for `get_winding`, `all_true`, `any_true`

petersalemink95 (changes requested):

petersalemink95:
Can we put the filters in `filters/_flters.py` and then import the functions in the `__init__`?
I don't think we should define the functions within the` __init__ ` file

src/power_grid_model_io/filters/__init__.py
bramstoeller: I don't think so. I guess it's rudimentary code, written when z-winding wasn't supported yet. @TonyXiang8787 Can we remove this if statement (and thereby the neutral_grounding parameter)?

tests/unit/filters/test_filters.py
bramstoeller: Noted. I'll add them.

bramstoeller:
> Can we put the filters in `filters/_flters.py` and then import the functions in the `__init__`? I don't think we should define the functions within the`__init__` file

Yes, done.

src/power_grid_model_io/filters/__init__.py
TonyXiang8787: It is possible to have winding `wye_n` but neutral is not grounded.

So we need this logic.

bramstoeller: ok

tests/unit/filters/test_filters.py
bramstoeller: Done.

PR #58, opened 2022-11-03
Auto ids (for Vision files)

bramstoeller:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -227,27 +288,43 @@ grid:
...
     sym_gen:
-      - id: #10
-          - Node.Number
-          - Subnumber
-          - 10
+      - id:
+          auto_id:
+            name:
+              - Transformer loads
+              - generation
```
bramstoeller, line 295: generator?

petersalemink95 (approved):

src/power_grid_model_io/config/excel/vision_en.yaml
petersalemink95: What's it called in Vision? I think we need to keep the names similar

bramstoeller: Well, we do have the columns named `Generation.P` and `Generation.Q`, so let's keep `generation`.

PR #60, opened 2022-11-08
Panda Power to PGM converter

Laurynas-Jagutis:
https://github.com/alliander-opensource/power-grid-model-io/issues/1


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+
+        self.pgm_data["line"] = pgm_lines
+
+    def _create_pgm_input_sources(self):
+        assert "source" not in self.pgm_data
+
+        pp_ext_grid = self.pp_data["ext_grid"]
+
+        pgm_sources = initialize_array(data_type="input", component_type="source", shape=len(pp_ext_grid))
+        pgm_sources["id"] = self._generate_ids("source", pp_ext_grid.index)
```
nitbharambe, line 151: I think the ids would be generated by the `table name` ie. names of components from pandapower. Then it be `ext_grid` here and similar in other functions as well.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+
+        n_loads = len(pp_loads)
+
+        pgm_sym_loads = initialize_array(data_type="input", component_type="sym_load", shape=3 * n_loads)
+
+        const_i_multiplier = pp_loads["const_i_percent"] * (1e-2 * 1e6)
+        const_z_multiplier = pp_loads["const_z_percent"] * (1e-2 * 1e6)
+        const_p_multiplier = 1e6 - const_i_multiplier - const_z_multiplier
+
+        pgm_sym_loads["id"][:n_loads] = self._generate_ids("sym_load_const_power", pp_loads.index)
```
nitbharambe, line 199: Something more must be thought in generating ids of sym_load for recombining results

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def _create_pgm_input_transformers(self):
+        assert "transformer" not in self.pgm_data
+
+        pp_trafo = self.pp_data["trafo"]
+
+        pgm_transformers = initialize_array(data_type="input", component_type="transformer", shape=len(pp_trafo))
+        pgm_transformers["id"] = self._generate_ids("transformer", pp_trafo.index)
+        pgm_transformers["from_node"] = self._get_ids("bus", pp_trafo["hv_bus"])
+        pgm_transformers["from_status"] = pp_trafo["in_service"]
```
nitbharambe, line 230: Should add switches here similar to line.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+        pgm_transformers["to_node"] = self._get_ids("bus", pp_trafo["lv_bus"])
+        pgm_transformers["to_status"] = pp_trafo["in_service"]
+        pgm_transformers["u1"] = pp_trafo["vn_hv_kv"] * 1e3
+        pgm_transformers["u2"] = pp_trafo["vn_lv_kv"] * 1e3
+        pgm_transformers["sn"] = pp_trafo["sn_mva"] * 1e6
+        pgm_transformers["uk"] = pp_trafo["vkr_percent"] * 1e-2
```
nitbharambe, line 236: `uk` must be mapped with `vk_percent`

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+        pgm_transformers["pk"] = (pp_trafo["vkr_percent"] * 1e-2) * (pp_trafo["sn_mva"] * 1e6)
+        pgm_transformers["i0"] = pp_trafo["i0_percent"] * 1e-2
+        pgm_transformers["p0"] = pp_trafo["pfe_kw"] * 1e3
+        pgm_transformers["winding_from"] = pp_trafo["vector_group"].apply(get_transformer_winding_from)
+        pgm_transformers["winding_to"] = pp_trafo["vector_group"].apply(get_transformer_winding_to)
+        pgm_transformers["clock"] = (pp_trafo["shift_degree"] / 30).astype(np.int8)
```
nitbharambe, line 242: We should also warn user in case of non-clock shift degree values. ie. shift degree values not multiples of 30

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+        self.idx_lookup[key] = pd.Series(pp_idx, index=pgm_idx)
+        self.next_idx += n_objects
+        return pgm_idx
+
+    def _get_ids(self, key: str, pp_idx: pd.Series) -> pd.Series:
+        if key not in self.idx:
+            raise KeyError(f"No indexes have been created for '{key}'!")
+        return self.idx[key][pp_idx]
+
+    @staticmethod
```
nitbharambe, line 319: Just curious: Why use static method here?

Apply function processes a single row. And maybe its possible for this function to be in vectorized form.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,361 @@
...
+    def _get_tap_size(pp_trafo: pd.DataFrame) -> np.ndarray:
+        tap_side_hv = np.array(pp_trafo["tap_side"] == "hv")
+        tap_side_lv = ~tap_side_hv  # everything else that is not hv is lv
```
nitbharambe, line 322: Invalid values can appear too, we must warn user in that case.

src/power_grid_model_io/converters/pandapower_converter.py
Laurynas-Jagutis: Yes, that is true, thanks! I assume that we can't do that with loads, because we create three of them. I do leave them as "sym_load_const_power", "sym_load_const_impedance", etc. Right?

Laurynas-Jagutis: Oh, your next comment regards loads

src/power_grid_model_io/converters/pandapower_converter.py
Laurynas-Jagutis: Could you explain the issue or maybe give an idea on how to solve it?

src/power_grid_model_io/converters/pandapower_converter.py
Laurynas-Jagutis: I should add them to three winding transformer as well then, right?

Laurynas-Jagutis: Added them to transformer and three winding transformer.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,460 @@
...
+        pgm_transformers["to_node"] = self._get_ids("bus", pp_trafo["lv_bus"])
+        pgm_transformers["to_status"] = pp_trafo["in_service"] & pp_to_switches["closed"]
+        pgm_transformers["u1"] = pp_trafo["vn_hv_kv"] * 1e3
+        pgm_transformers["u2"] = pp_trafo["vn_lv_kv"] * 1e3
+        pgm_transformers["sn"] = pp_trafo["sn_mva"] * 1e6
+        pgm_transformers["uk"] = pp_trafo["vk_percent"] * 1e-2
+        pgm_transformers["pk"] = (pp_trafo["vkr_percent"] * 1e-2) * (pp_trafo["sn_mva"] * 1e6)
+        pgm_transformers["i0"] = pp_trafo["i0_percent"] * 1e-2
+        pgm_transformers["p0"] = pp_trafo["pfe_kw"] * 1e3
+        pgm_transformers["winding_from"] = pp_trafo["vector_group"].apply(get_transformer_winding_from)
```
bramstoeller, line 273: I'm reviewing using the `pp_power_flow.py` example script. Using that script, `pp_trafo` doesn't contain `vector_group`. It does contain `std_type`, so I guess we need to load the `vector group` from the std types.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,460 @@
...
+        # Add an extra column called 'index' containing the index values, which we need to do the join with switches
+        pp_trafo["index"] = pp_trafo.index
+
+        # Select the appropriate switches and columns
+        pp_switches = self.pp_data["switch"]
+        pp_switches = pp_switches[self.pp_data["switch"]["et"] == "t"]
+        pp_switches = pp_switches[["element", "bus", "closed"]]
+
+        # Join the switches with the trafo twice, once for the from_bus and once for the to_bus
+        pp_from_switches = (
```
bramstoeller, line 237: Would be nice to write a function:
```python
def get_switch_states(component: pd.DataFrame, switches: pd.DataFrame, bus: str) -> pd.Series:
    ...
```

By creating a function, you reduce the risks for errors and you can test the merge logic inj a unit test.
It also make it easier to profile the function later.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,460 @@
...
+    def _create_pgm_input_shunts(self):
+        assert "shunt" not in self.pgm_data
+
+        pp_shunt = self.pp_data["shunt"]
+
+        pgm_shunts = initialize_array(data_type="input", component_type="shunt", shape=len(pp_shunt))
+        pgm_shunts["id"] = self._generate_ids("shunt", pp_shunt.index)
+        pgm_shunts["node"] = self._get_ids("bus", pp_shunt["bus"])
+        pgm_shunts["status"] = pp_shunt["in_service"]
+        pgm_shunts["g1"] = (pp_shunt["p_mw"] * 1e6) / ((pp_shunt["vn_kv"] * 1e3) * (pp_shunt["vn_kv"] * 1e3))
```
bramstoeller, line 168: You could combine the multiplications. Right now you have

```python
g = p * 1e6 / ((v * 1e3) * (v * 1e3))
b = q * 1e6 / ((v * 1e3) * (v * 1e3))
```
Which is 10 x N multiplications, while this is the same as

```python
g = p * 1e6 / (v * v * 1e6)
b = q * 1e6 / (v * v * 1e6)
```
Which is 8 x N multiplications, while this is the same as
```python
g = p * v * v
b = q * v * v
```
Which is 4 x N multiplications, while this is the same as
```python
v2 = v * v
g = p * v2
b = q * v2
```
Which is 3 x N multiplications

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,460 @@
...
+        pgm_3wtransformers["u2"] = pp_trafo3w["vn_mv_kv"] * 1e3
+        pgm_3wtransformers["u3"] = pp_trafo3w["vn_lv_kv"] * 1e3
+        pgm_3wtransformers["sn_1"] = pp_trafo3w["sn_hv_mva"] * 1e6
+        pgm_3wtransformers["sn_2"] = pp_trafo3w["sn_mv_mva"] * 1e6
+        pgm_3wtransformers["sn_3"] = pp_trafo3w["sn_lv_mva"] * 1e6
+        pgm_3wtransformers["uk_12"] = pp_trafo3w["vk_hv_percent"] * 1e-2
+        pgm_3wtransformers["uk_13"] = pp_trafo3w["vk_lv_percent"] * 1e-2
+        pgm_3wtransformers["uk_23"] = pp_trafo3w["vk_mv_percent"] * 1e-2
+
+        pgm_3wtransformers["pk_12"] = (pp_trafo3w["vkr_hv_percent"] * 1e-2) * (
```
bramstoeller, line 353: You could combine the multiplications. Right now you have

```python
pk = (vkr * 1e-2) * (min(sn_hv, sn_mv) * 1e6)
```
Which is 3 x N multiplications, while this is the same as
```python
pk = vkr * min(sn_hv, sn_mv) * (1e-2 * 1e6)
```
Which is 2 x N + 1 multiplications

bramstoeller (commented):
Structure seems fine. Some remarks on efficiency. We'll dive into more optimizations later.

src/power_grid_model_io/converters/pandapower_converter.py
Laurynas-Jagutis: Thank you for such a clear explanation!
A quick question, shouldn't this be division? As in 
g = p / v2
b = q / v2

src/power_grid_model_io/converters/pandapower_converter.py
nitbharambe: Just that it needs to be thought of while making output conversions. I think you have it covered here.

docs/examples/pp_power_flow.py
```diff
@@ -0,0 +1,43 @@
...
+
+from power_grid_model_io.converters.pandapower_converter import PandaPowerConverter
+from power_grid_model_io.converters.pgm_json_converter import PgmJsonConverter
+
+structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(logging.DEBUG))
+
+# Source data
+pp_net = pandapower.networks.example_simple()
+
+pp_data = {component: pp_net[component] for component in pp_net if isinstance(pp_net[component], pd.DataFrame)}
```
nitbharambe, line 21: We also need `pp_data.f_hz` now which is a float.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,518 @@
...
+        pgm_lines["from_status"] = pp_lines["in_service"] & switch_states.iloc[0, :]
+        pgm_lines["to_node"] = self._get_ids("bus", pp_lines["to_bus"])
+        pgm_lines["to_status"] = pp_lines["in_service"] & switch_states.iloc[1, :]
+        pgm_lines["r1"] = pp_lines["r_ohm_per_km"] * pp_lines["length_km"] / pp_lines["parallel"]
+        pgm_lines["x1"] = pp_lines["x_ohm_per_km"] * pp_lines["length_km"] / pp_lines["parallel"]
+        pgm_lines["c1"] = pp_lines["c_nf_per_km"] * pp_lines["length_km"] * pp_lines["parallel"] * 1e-9
+
+        # The formula for tan1 = R_1 / Xc_1 = (g * 1e-6) / (2 * pi * f * c * 1e-9) = g / (2 * pi * f * c * 1e-3)
+        pgm_lines["tan1"] = pp_lines["g_us_per_km"] / (
+            2 * np.pi * self.system_frequency * pp_lines["c_nf_per_km"] * 1e-3
```
TonyXiang8787, line 110: put all the coefficient at the front. So they calculate one scaling factor and then multiply the whole vector.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,518 @@
...
+    def _create_pgm_input_sources(self):
+        assert "source" not in self.pgm_data
+
+        pp_ext_grid = self.pp_data["ext_grid"]
+
+        pgm_sources = initialize_array(data_type="input", component_type="source", shape=len(pp_ext_grid))
+        pgm_sources["id"] = self._generate_ids("ext_grid", pp_ext_grid.index)
+        pgm_sources["node"] = self._get_ids("bus", pp_ext_grid["bus"])
+        pgm_sources["status"] = pp_ext_grid["in_service"]
+        pgm_sources["u_ref"] = pp_ext_grid["vm_pu"]
```
TonyXiang8787, line 126: Also convert `va_degree` for `u_angle_ref` in pgm.

Also convert `rx_ratio` and `z01_ratio` in pgm.

.pre-commit-config.yaml
```diff
@@ -34,3 +34,4 @@ repos:
         language: system
         pass_filenames: false
         always_run: true
+        args: [ "--cov-fail-under=95" ]
```
petersalemink95, line 37: 1. Do we need to specify it here, since it is already in pyproject.toml? 
2. Why is it 95? In the pyproject.toml on main we have 100

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,507 @@
...
+import pandas as pd
+from power_grid_model import Branch3Side, BranchSide, LoadGenType, initialize_array
+from power_grid_model.data_types import Dataset
+
+from power_grid_model_io.converters.base_converter import BaseConverter
+from power_grid_model_io.data_types import ExtraInfoLookup
+from power_grid_model_io.filters import get_winding
+
+StdTypes = Mapping[str, Mapping[str, Mapping[str, Union[float, int, str]]]]
+PandasData = Mapping[str, pd.DataFrame]
```
petersalemink95, line 21: I see that it is often called Pandas data, do we want this? It is indeed just a Pandas dataframe, but don't we want to call it PandaPowerData here? Since this module will only be used for pandapower and not for other Pandas DataFrames

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -0,0 +1,507 @@
...
+        self.idx_lookup = {}
+        self.next_idx = 0
+
+        # Set pandas data
+        self.pp_data = data
+        # Convert
+        if data_type == "input":
+            self._create_input_data()
+        else:
+            raise NotImplementedError()
```
petersalemink95, line 57: Do we want this to be more specific? e.g. something like:
`raise NotImplementedError(f"Data type: '{data_type}' is not implemented")`

.pre-commit-config.yaml
bramstoeller: While developing you may want to have a lower coverage, so you can easily commit intermediate versions.

src/power_grid_model_io/converters/pandapower_converter.py
bramstoeller: It's not 'just' pandas data, it is a dictionary of str to DataFrame. But indeed, PandaPowerData may be better as it is only used in this file.

src/power_grid_model_io/converters/pandapower_converter.py
bramstoeller: Yes, that would be better. Or maybe we should raise a `ValueError` in stead. I don't think we will implement other data types in the future.

.pre-commit-config.yaml
petersalemink95: Makes sense :)

src/power_grid_model_io/converters/pandapower_converter.py
petersalemink95: Yes, I think `ValueError` would be better indeed

docs/examples/pp_power_flow.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model IO project <dynamic.grid.calculation@alliander.com>
```
bramstoeller, line 1: Shall we remove this file and only keep the notebook? I like example scripts for development and debugging, but for production (main branch) we chose notebooks.

petersalemink95: I agree, make sure the same information is also provied in the notebook

docs/examples/pandapower_example.ipynb
```diff
@@ -0,0 +1,634 @@
...
+    "\n",
+    "This example illustrates conversion from PandaPower to power-grid-model input data. \n",
+    "We can then calculate power-flow with it or convert to a different formats like PGM JSON."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. Load the PandaPower Data\n",
```
bramstoeller, line 17: The notebook is not working. I will create a new branch to propose some fixes.

bramstoeller (changes requested):
Looks very good, but make sure that we handle all possible PandaPower networks. For example, the `pandapower.networks.example_simple()` can't be converted using the current implementation.

I don't know the in's and out's of the std_types and optional attributes in pandapower, but maybe instead of `self.pp_data[table][attr]` we should use the `.get` function, like `self.pp_data[table].get(attr, np.nan)`. Or even a member function like `self._get_pp_attr(table, attr, np.nan)`, which could be implemented something like this (not tested):

```python
def _get_pp_attr(self, table: str, attribute: str, default: Optional[float] = None) \
        -> Union[np.ndarray, float]:
    pp_data = self.pp_data[table]

    # If the attribute exists, return it
    if attribute in pp_data:
        return pp_data[attribute]

    # Try to find the std_type value for this attribute
    if table in self.std_types[table] and "std_type" in pp_data:
        std_types = self.std_types[table]

        @lru_cache
        def get_std_value(std_type_name: str):
            std_type = std_types[std_type_name]
            if attribute in std_type:
                return std_types[attribute]
            if default is not None:
                return default
            raise KeyError(f"No '{attribute}' value for '{table}' with std_type '{std_type_name}'.")

        if self.std_types is not None and table in self.std_types:
            return pp_data.apply(get_std_value)

    # Return the default value (assume that broadcasting is handled by the caller / numpy)
    if default is None:
        raise KeyError(f"No '{attribute}' value for '{table}'.")
    return default
```

docs/examples/pandapower_example.ipynb
nitbharambe: @Laurynas-Jagutis is already taking care of the get function for attributes. But lets merge only the notebook.

nitbharambe (approved):

bramstoeller (approved):

PR #65, opened 2022-11-10
validation export file for vision

nitbharambe:


bramstoeller (approved):
Could you supply the corresponding pgm .json file as well?

tests/data/vision/vision_validation.json
```diff
@@ -0,0 +1,46 @@
+{
```
bramstoeller, line 1: Did you manually check the values?

bramstoeller (approved):
Sorry, could you add the source file (.vnf) as well?

PR #69, opened 2022-11-16
Feature/auto id dict

bramstoeller:


src/power_grid_model_io/converters/gaia_excel_converter.py
```diff
@@ -8,13 +8,13 @@
...
 
-from power_grid_model_io.converters.tabular_converter import TabularConverter
+from power_grid_model_io.converters.vision_excel_converter import VisionExcelConverter
 from power_grid_model_io.data_stores.gaia_excel_file_store import GaiaExcelFileStore
 
 DEFAULT_MAPPING_FILE = Path(__file__).parent.parent / "config" / "excel" / "gaia_{language:s}.yaml"
 
 
-class GaiaExcelConverter(TabularConverter):
+class GaiaExcelConverter(VisionExcelConverter):
```
petersalemink95, line 17: Do we want this kind of inheritance? i.e. Base-Tabular-VisionExcel-GaiaExcel

Seems to me that you'd want Vision and Gaia both to be final child classes. It would maybe create a bit more overhead, but it does not seem logical to me to let Gaia inherit from Vision, since the two programs are completely different.

Then you would need an `ExcelConverter` class in between and get:
Base-Tabular-Excel-VisionExcel
and 
Base-Tabular-Excel-GaiaExcel

What do you think?

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -511,18 +510,29 @@ def _parse_col_def_filter(self, data: TabularData, table: str, col_def: Dict[str
         return pd.concat(data_frames, axis=1)
 
     def _parse_auto_id(
-        self, data: TabularData, table: str, name: Union[str, List[str]], key_col_def: Union[str, List[str]]
+        self,
```
petersalemink95, line 513: Could you add a docstring about the functionality?

src/power_grid_model_io/converters/gaia_excel_converter.py
petersalemink95: Agreed

petersalemink95 (approved):

PR #74, opened 2022-11-17
Restore extra info

bramstoeller:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -137,14 +146,15 @@ grid:
       r_grounding_to: Re2
       x_grounding_to: Xe2
       extra:
-        - Name
+        - Number
```
petersalemink95, line 149: transformer used to have Subnumber

petersalemink95: (and Node.Number)

src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -244,8 +255,10 @@ grid:
       r_grounding_to: 0
       x_grounding_to: 0
       extra:
-        - Name
+        - Node.Number
```
petersalemink95, line 258: I guess the same here (Node.Number / Subnumber)

src/power_grid_model_io/config/excel/vision_en.yaml
bramstoeller: I think that's only for transformers created from the Transformer loads table.

src/power_grid_model_io/config/excel/vision_en.yaml
bramstoeller: Whet do you mean? Here it does say Node.Number and Subnumber, right?

petersalemink95 (approved):

PR #75, opened 2022-11-17
Validation test for extra info

bramstoeller:


tests/data/vision/vision_validation.json
```diff
@@ -9,7 +9,7 @@
     ],
   "line":
     [
-      {"id": 4, "from_node": 0, "to_node": 1},
+      {"id": 4, "from_node": 0, "to_node": 1, "Name": "cable1"},
```
petersalemink95, line 12: Why is this the only component for which "Name" is given in the json?

bramstoeller: Because I was implementing the mechanism, not the actual unit test. We could make that part of this issue though.
What do you think? Should we do that?

petersalemink95: I guess it would be best to do it here. We could make another issue, but I think it would be best to test it immediately

petersalemink95 (approved):

PR #81, opened 2022-11-23
New id_reference structure

bramstoeller:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -7,7 +7,6 @@ grid:
     node:
       id:
         auto_id:
-          name: Nodes
```
petersalemink95, line 10: `table: Nodes` should still be here I think?

petersalemink95 (approved):

PR #99, opened 2022-12-08
Feature/explicit mapping

bramstoeller:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -549,50 +571,50 @@ grid:
...
           - Scaling
       extra:
         - ID
         - Name
 units:
-  A:
+  A: null
   F:
-    µF: 0.000_001
+    µF: 0.000001
```
petersalemink95, line 594: Underscores are gone and values changed to int

bramstoeller: whoops... thnx, just fixed it.

docs/converters/tabular_converter.md
```diff
@@ -173,3 +189,117 @@ item = auto_id[1]      # item = "Bravo"
...
+        to_node:
+          auto_id:
+            table: Nodes
+            key:
+              Number: To_Number
+```
+This basically reads as:
+* For each row in the Nodes table, a PGM node instance is created.
+  * For each node instance, a numerical id is generated, which is unique for each value in the Number column. This 
+    assumes that the Number column is unique in the source table. Let's say tha values of the Number column in the
```
petersalemink95, line 221: that*

petersalemink95 (approved):

TonyXiang8787 (approved):

PR #103, opened 2022-12-13
pp to pgm v2 fixes

bramstoeller:


bramstoeller:
This branch is just a suggestion. If this suggestion is to be accepted, there are probably many more changes to be made, unit tests to be added and some cases to be added to the pp_to_pgm 'unit' tests (which is actually validation test).

nitbharambe:
@bramstoeller The notebook still needs to be updated right?

PR #109, opened 2022-12-20
Conversion for asymmetrical load from PP to PGM

Laurynas-Jagutis:


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -242,6 +243,35 @@ def _create_pgm_input_sym_loads(self):
...
+
+        scaling = self._get_pp_attr("asymmetric_load", "scaling")
+        multiplier = 1e6 * scaling
+
+        pgm_asym_loads = initialize_array(data_type="input", component_type="asym_load", shape=len(pp_asym_loads))
+        pgm_asym_loads["id"] = self._generate_ids("asymmetric_load", pp_asym_loads.index)
+        pgm_asym_loads["node"] = self._get_ids("bus", pp_asym_loads["bus"])
+        pgm_asym_loads["status"] = self._get_pp_attr("asymmetric_load", "in_service")
+        pgm_asym_loads["p_specified"] = np.array(
+            self._get_pp_attr("asymmetric_load", "p_a_mw") * multiplier,
```
bramstoeller, line 262: It might be slightly faster to apply the multiplier just once, to the numpy array. It will probably be unmeasurable, but for the consistency in code style it would be better.

bramstoeller:
```python
pgm_asym_loads["p_specified"] = np.array(
            self._get_pp_attr("asymmetric_load", "p_a_mw"),
            self._get_pp_attr("asymmetric_load", "p_b_mw"),
            self._get_pp_attr("asymmetric_load", "p_c_mw"),
        ) * multiplier
```

bramstoeller (dismissed):
I had one remark, just change it if you agree, or ignore it if you don't agree :)

src/power_grid_model_io/converters/pandapower_converter.py
Laurynas-Jagutis: Updated the conversion with your suggestion.

bramstoeller (approved):

nitbharambe (approved):

PR #113, opened 2022-12-21
Included docstrings

Laurynas-Jagutis:
Signed-off-by: Laurynas Jagutis <laurynas.jagutis@alliander.com>


bramstoeller (approved):
Very nice! I'm not really sure this is needed for private functions, but it is always good to have it. Especially if other people (from pp?) may contribute or want to debug in the future.

nitbharambe (approved):

PR #117, opened 2023-01-05
Move regular expressions to utils

bramstoeller:
... and add unit tests

### DIFF
--- a/src/power_grid_model_io/utils/regex.py
+++ b/src/power_grid_model_io/utils/regex.py
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+"""
+General regular expressions
+"""
+
+import re
+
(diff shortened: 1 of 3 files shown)


src/power_grid_model_io/utils/regex.py
```diff
@@ -0,0 +1,42 @@
...
+
+import re
+
+TRAFO_CONNECTION_RE = re.compile(r"^(Y|YN|D|Z|ZN)(y|yn|d|z|zn)(\d|1[0-2])?$")
+r"""
+Regular expressions to the winding_from and winding_to codes and optionally the clock number:
+^               Start of the string
+(Y|YN|D|Z|ZN)   From winding type
+(y|yn|d|z|zn)   To winding type
+(\d|1[0-2])?    Optional clock number (0-12)
```
nitbharambe, line 16: `vector_group` attriubte of pandapower does not have clock number in the string. The `shift_degree` attribute gets the clock.

nitbharambe (approved):

src/power_grid_model_io/utils/regex.py
bramstoeller: That's why it's optional, to follow the notation in vision.

nitbharambe: > That's why it's optional, to follow the notation in vision. Both Vision and pandapower are using the same pattern now.

Makes sense!

PR #132, opened 2023-01-17
Pandapower asymmetrical input and output conversion

Laurynas-Jagutis:
Signed-off-by: Laurynas Jagutis <laurynas.jagutis@alliander.com>


tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -638,6 +705,38 @@ def test_create_pgm_input_asym_loads(mock_init_array: MagicMock, two_pp_objs, co
     assert converter.pgm_input_data["asym_load"] == mock_init_array.return_value
 
 
+def test_create_pgm_input_sym_loads__asym() -> None:
```
petersalemink95, line 708: Only delta loads are tested here. Are wye loads tested somewhere else? If so, I think this function should have a more meaningful name, e.g. `test_create_pgm_input_sym_loads__asym_delta` or something like that. If not, shouldn't that be implemented?

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -638,6 +705,38 @@ def test_create_pgm_input_asym_loads(mock_init_array: MagicMock, two_pp_objs, co
...
+    converter.pp_input_data = {k: v for k, v in pp_net.items() if isinstance(v, pd.DataFrame)}
+
+    # Act/Assert
+    with pytest.raises(
+        NotImplementedError, match="Delta loads are not implemented, only wye loads are supported in PGM."
+    ):
+        converter._create_pgm_input_sym_loads()
+
+
+def test_create_pgm_input_asym_loads__asym() -> None:
```
petersalemink95, line 724: Same question here

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -65,6 +96,21 @@ def test_generate_output():  # TODO: REMOVE THIS FUNCTION
     json_converter.save(data=output_data, extra_info=extra_info)
 
 
+def test_generate_output_3ph():  # TODO: REMOVE THIS FUNCTION
```
petersalemink95, line 99: When or why should this function be removed?

nitbharambe: We shall remove this in the clearing of todos task

tests/unit/converters/test_pandapower_converter_input.py
nitbharambe: Renamed

petersalemink95 (approved):

PR #136, opened 2023-01-27
Fixes post big vision network validation

nitbharambe:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -33,8 +33,12 @@ grid:
...
       c1: C
       tan1: 0
-      r0: R0
-      x0: X0
+      r0:
+        power_grid_model_io.functions.zeros_to_nan:
+          value: R0
+      x0:
+        power_grid_model_io.functions.zeros_to_nan:
+          value: X0
```
TonyXiang8787, line 41: This is not correct. We cannot always convert zeros to NaN.

You can have a valid `r0=0` and `x0=1`.

src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -62,8 +66,12 @@ grid:
       x1: X
       c1: C
       tan1: 0
-      r0: R0
-      x0: X0
+      r0:
+        power_grid_model_io.functions.zeros_to_nan:
+          value: R0
```
TonyXiang8787, line 71: same here

src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -111,8 +119,12 @@ grid:
       x1: X
       c1: 0
       tan1: 0
-      r0: R0
-      x0: X0
+      r0:
+        power_grid_model_io.functions.zeros_to_nan:
+          value: R0
```
TonyXiang8787, line 124: same here

src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -588,6 +666,111 @@ grid:
...
+      tap_min: Tap min a
+      tap_max: Tap max a
+      tap_nom: Tap nom a
+      tap_size: Tap size a
+      r_grounding_1: Re1
+      x_grounding_1: Xe1
+      r_grounding_2: Re2
+      x_grounding_2: Xe2
+      r_grounding_3: Re3
+      x_grounding_3: Xe3
```
TonyXiang8787, line 769: Do you also have a test case for three winding transformer?

nitbharambe: Yes, there is a three winding transformer in the validation case

TonyXiang8787: Why did the validation test not fail when we did not have conversion for three-winding transformer yet?

bramstoeller: I'll have a look...

bramstoeller: Three winding transformer was not in the validation json file. Only the components and attributes available in the json file are validated. But I still don't  see them in the json file for this branch...

Further more, 'not all sheets contain components. So sheets that are not included in the yaml are considered not to contain components. We could fix this by adding all the sheets that do not contain components to the yaml file as well and raise an exception  if a sheet is found in the xlsx file that is not defined in the yaml file.

TonyXiang8787: > Furthermore, not all sheets contain components. So sheets that are not included in the yaml are considered not to contain components. We could fix this by adding all the sheets that do not contain components to the yaml file as well, and raise an exception if a sheet is found in the xlsx file that is not defined in the yaml file.

@bramstoeller I am hesitating on this one. Phase to Phase adds new sheets in the Excel file regularly. If we do this, the conversion will be breaking too often.

nitbharambe: > Three winding transformer was not in the validation json file. Only the components and attributes available in the json file are validated. But I still don't see them in the json file for this branch...
> 
> Furthermore, not all sheets contain components. So sheets that are not included in the yaml are considered not to contain components. We could fix this by adding all the sheets that do not contain components to the yaml file as well, and raise an exception if a sheet is found in the xlsx file that is not defined in the yaml file.

I added the component in json now. I am now updating the nl language file too.

bramstoeller (changes requested):
I've checked the python code and unit tests, but not the YAML file, as I don't know the correct formulas. Code and unit tests seem ok.

We are missing some validation tests. That is, I do see the input data (e.g. three winding transformer) in the xlsx file, but there is nor output data to validate against in the json file. Could you please add those and include a three winding transformer that highlights the case that Tony mentioned in his review.

nitbharambe:
> I've checked the python code and unit tests, but not the YAML file, as I don't know the correct formulas. Code and unit tests seem ok.
> 
> We are missing some validation tests. That is, I do see the input data (e.g. three winding transformer) in the xlsx file, but there is nor output data to validate against in the json file. Could you please add those and include a three winding transformer that highlights the case that Tony mentioned in his review.

Added the three winding transformer in json. There are only validation tests for conversion and not outputs right?

src/power_grid_model_io/config/excel/vision_en.yaml
nitbharambe: Right! Modified the function to align that way.

TonyXiang8787 (approved):

PR #148, opened 2023-02-13
Bug/143 vision excel converter only english

pkuppens:
1. I tried to create a solution that uses the global words for tables, nodes, etc. in the business domain while creating a mapping to local words for a converter so it should work in "nl" with the Dutch words.
2. Added a few tests
3. I noticed that the Sonar build failed in the Actions tab in my fork, do I need something like a SONAR_TOKEN? Is this something that would work in the alliander-opensource repo?
4. Code is a 'my-first-pull-request' on this repo, so be careful and polite ;)


nitbharambe:
Hi Pieter,
Thanks for finding the bug! And more so for suggesting a good fix. 
Also @bramstoeller , can you mention the comments from our discussion here?

nitbharambe:
I think it would okay to have the DEFAULT_MAPPING_KEYS as a separate yaml file than the python one. :)
Thanks again by the way!

bramstoeller:
Hi Pieter,

Thanks for your contribution! I like the idea, but I feel that it might be cleaner to add something like this to the yaml files:
```yaml
references:
    node_table: Knooppunten
    number: Nummer
    node_number: Knooppunt.Nummer
    sub_number: Subnummer
```

Furthermore, I don't think we should translate thinks like Capacitors/Condensatoren. The `table` argument refers to the sheet name in the excel export. If the sheet name was Dutch, then the user should provide the Dutch name to the function as well. It would be confusing for many data scientists (at least at Alliander) to have to use the English names. If someone would like to use the english names, they should make english Excel exports.

src/power_grid_model_io/converters/vision_excel_converter.py
```diff
@@ -20,32 +141,57 @@ class VisionExcelConverter(TabularConverter):
     """
 
     def __init__(self, source_file: Optional[Union[Path, str]] = None, language: str = "en"):
-        mapping_file = Path(str(DEFAULT_MAPPING_FILE).format(language=language))
+        mapping_file = Path(str(DEFAULT_MAPPING_FILE).format(language=str(language)))
```
bramstoeller, line 144: Interesting, why do you use the `str()` function? Is it to make sure that language is a string? I think we shouldn't (have to) do that. It's up to the user to supply the right arguments, in the correct data type.

tests/unit/converters/test_vision_excel_converter.py
```diff
@@ -32,22 +32,59 @@ def converter() -> VisionExcelConverter:
...
+
+
+@pytest.mark.parametrize("language", ["en", "nl"])
 def test_mapping_files_exist(language: str):
     vf = Path(str(DEFAULT_MAPPING_FILE).format(language=language))
     assert vf.exists()
 
 
-def test_initialization():
+def test_mapping_file_none_input_raises_file_not_found_error():
```
bramstoeller, line 64: `language` should be a string, so there we don't have to test this. It's up to the user to supply to respect the type hints.

tests/unit/converters/test_vision_excel_converter.py
```diff
@@ -32,22 +32,59 @@ def converter() -> VisionExcelConverter:
...
+    with pytest.raises(FileNotFoundError, match="No Vision Excel mapping available for language 'None'"):
+        VisionExcelConverter(language=None)
+
+
+def test_mapping_file_numeric_input_raises_file_not_found_error():
```
bramstoeller, line 69: `language` should be a string, so there we don't have to test this. It's up to the user to supply to respect the type hints.

tests/unit/converters/test_vision_excel_converter.py
```diff
@@ -32,22 +32,59 @@ def converter() -> VisionExcelConverter:
...
+    with pytest.raises(FileNotFoundError, match="No Vision Excel mapping available for language '1'"):
+        VisionExcelConverter(language=1)
+
+
+def test_mapping_file_not_exists_raises_file_not_found_error():
```
bramstoeller, line 74: Very good idea to split these tests!

tests/unit/converters/test_vision_excel_converter.py
pkuppens: The reasoning is that the language can be misread from a configuration file/environment setting or so, and evaluated to None.
In such a case, this will clarify the error message.

bramstoeller: Ok, fair enough. But than I'd rather check it using:
```python
assert isinstance(language, str)
```
or
```python
if not isinstance(language, str):
    raise TypeError(f"`language` should be a string value, not {type(langue).__name__}.")
```
In this case I prefer the first one, because at this point we're dealing with a programming error, rather than a runtime error. Someone called the function with the wrong type, so we don't have to be polite and explain exactly what he/she did wrong. The assert is shorter and has 'less complexity' (according to sonar etc).

bramstoeller:
@pkuppens, I just had a discussion with the team and with @jaapschoutenalliander, the reporter of the bug. We have decided that the user should use the Dutch table names if the export was in Dutch. So only the words 'Nodes', 'Number', 'Node.Number' and 'Subnumber' should be translated and the translation should be supplied in the mapping file.

bramstoeller:
@pkuppens , Thank you for your contribution. Your suggestions will partially be implemented in https://github.com/alliander-opensource/power-grid-model-io/pull/175

PR #150, opened 2023-02-14
Add default tap and vector group for pandapower converter

nitbharambe:


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -535,11 +535,20 @@ def _create_pgm_input_transformers(self):
...
         sn_mva = self._get_pp_attr("trafo", "sn_mva")
         switch_states = self.get_switch_states("trafo")
-        winding_types = self.get_trafo_winding_types()
 
-        tap_side = self._get_pp_attr("trafo", "tap_side")
+        tap_side = self._get_pp_attr("trafo", "tap_side", np.nan)
         tap_nom = self._get_pp_attr("trafo", "tap_neutral", np.nan)
         tap_pos = self._get_pp_attr("trafo", "tap_pos", np.nan)
+        # Do not use taps when mandatory tap data is not available
+        no_taps = np.equal(tap_side, None) + np.isnan(tap_pos) + np.isnan(tap_nom)
```
TonyXiang8787, line 543: Use plus is confusing here. Consider use element wise or operator `|`. Or even use the function `np.logical_or`.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -532,11 +532,23 @@ def _create_pgm_input_transformers(self):
         parallel = self._get_pp_attr("trafo", "parallel", 1)
         sn_mva = self._get_pp_attr("trafo", "sn_mva")
         switch_states = self.get_switch_states("trafo")
-        winding_types = self.get_trafo_winding_types()
 
-        tap_side = self._get_pp_attr("trafo", "tap_side")
+        tap_side = self._get_pp_attr("trafo", "tap_side", np.nan)
```
TonyXiang8787, line 536: `tap_side` in `pandapower` is a string right? Why do you use the default value `np.nan` instead of `None`?

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -532,11 +532,23 @@ def _create_pgm_input_transformers(self):
...
+        tap_nom[no_taps] = 0
+        tap_pos[no_taps] = 0
+        tap_side[no_taps] = "hv"
+
+        winding_types = self.get_trafo_winding_types()
+        clocks = np.round(self._get_pp_attr("trafo", "shift_degree", 0.0) / 30) % 12
+        # Default vector group for odd clocks = DYn and for even clocks = YNyn
+        no_vector_groups = np.isnan(winding_types["winding_from"]) | np.isnan(winding_types["winding_to"])
+        no_vector_groups_dyn = no_vector_groups & (clocks % 2)
+        winding_types[no_vector_groups] = WindingType.wye_n
```
TonyXiang8787, line 550: Are you referring here to `winding_from` or `winding_to`?

nitbharambe: I set both, from and to windings to wye_n then only the odd clocks are set to delta

src/power_grid_model_io/converters/pandapower_converter.py
nitbharambe: Yes, its a string. Changed default to None

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -706,19 +708,19 @@ def test_create_pgm_input_transformers(mock_init_array: MagicMock, two_pp_objs,
...
-        pp_net, 0, 0, 0, 0, 0, 0, 0, 0, 0, tap_neutral=12.0, tap_pos=34.0, tap_side=None
-    )
+    args = [0, 0, 0, 0, 0, 0, 0, 0, 0]
+    pp.create_transformer_from_parameters(pp_net, *args, tap_neutral=12.0, tap_pos=34.0, tap_side="hv")
+    pp.create_transformer_from_parameters(pp_net, *args, tap_neutral=12.0, tap_pos=34.0, tap_side="lv")
+    pp.create_transformer_from_parameters(pp_net, *args, tap_neutral=12.0, tap_pos=34.0, tap_side=None)
+    tap_pos_trafo = pp.create_transformer_from_parameters(pp_net, *args, tap_neutral=12.0, tap_side="hv")
+    pp_net["trafo"]["tap_pos"][tap_pos_trafo] = np.nan
+    pp.create_transformer_from_parameters(pp_net, *args, tap_neutral=np.nan, tap_pos=34.0, tap_side="hv")
+    pp.create_transformer_from_parameters(pp_net, *args, vector_group=None, shift_degree=30)
```
petersalemink95, line 722: Since panda power doesn't require a shift degree in sets of 30 degrees I'd like to see a test with something in between, e.g. 48 degree to see if we handle this as expected

petersalemink95: Set the 30 -> 31 and the 60 -> 59, the results should still remain the same

nitbharambe: Added default test for clocks at arbitrary degree angles.

petersalemink95 (approved):

PR #151, opened 2023-02-15
Remove PP std types

bramstoeller:


nitbharambe (approved):

TonyXiang8787:
@bramstoeller we need to go through again with all components and attributes, and add default value if certain attributes are optional in pandapower.

PR #152, opened 2023-02-16
fix for no vector col in trafo

nitbharambe:
the get_winding_types did not have a return empty if not available like in `_get_pp_attr()`. Gives KeyError 
Fixed that.


tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -1565,8 +1572,13 @@ def test_get_trafo3w_winding_types__vector_group(mock_get_winding: MagicMock):
 
     expected = pd.DataFrame([[2, 1, 3], [1, 2, 0], [2, 1, 0]], columns=["winding_1", "winding_2", "winding_3"])
 
+    converter_empty = PandaPowerConverter()
```
bramstoeller, line 1575: It would be cleaner to create a new unit test for this. `test_get_trafo3w_winding_types__vector_group__missing`

bramstoeller (changes requested):
Please move the tests to separate test functions.

bramstoeller (approved):

PR #153, opened 2023-02-17
fixes #72

### LINKED ISSUE #72: Running mypy manually highlights multiple issues, while pre-commit reports none

pkuppens:
Replace type hints so mypy will not return an error:
1. numpy arange that is a function with the resulting type ndarray
2. numpy array that is a function with the resulting type ndarray
3. let mypy type checking ignore the numpy equal(attr_data, None) function


bramstoeller (commented):
Very well noticed, I don't know how this slipped into the repository...

sonar-project.properties
```diff
@@ -2,8 +2,10 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-sonar.projectKey=alliander-opensource_power-grid-model-io
-sonar.organization=alliander-opensource
+# sonar.projectKey=alliander-opensource_power-grid-model-io
```
bramstoeller, line 5: Hm, apparently this is an issue that we should solve... I don't know much about sonar... do you?
Or do you have an idea, @TonyXiang8787?

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -1621,7 +1621,7 @@ def _get_pp_attr(self, table: str, attribute: str, default: Optional[Union[float
         attr_data = pp_component_data[attribute]
 
         # If any of the attribute values are missing, and a default is supplied, fill the nans with the default value
-        nan_values = np.equal(attr_data, None) if attr_data.dtype is np.dtype("O") else np.isnan(attr_data)
+        nan_values = np.equal(attr_data, None) if attr_data.dtype is np.dtype("O") else np.isnan(attr_data)  # type: ignore
```
bramstoeller, line 1624: Do you know which type issue arises? Is there another way to solve it?

sonar-project.properties
bramstoeller: Sonar cloud should be skipped for external pull requests, so it should not be necessary to supply your own credentials.

src/power_grid_model_io/converters/pandapower_converter.py
pkuppens: To solve this, I have written 5 options in the original ticket.
For now, I think I implemented the simplest one, to let mypy ignore this single error.

pkuppens: @bramstoeller, this (sonar project properties) slips in the repo with automatic merges (instead of cherry picks).

One way to solve this issue is to ensure that you have the correct version in git, and then add it to the .gitignore, so updates won't get committed/pushed. (AFAIK, git will keep the original version).

bramstoeller (approved):

bramstoeller:
I've cherry picked your commit [da2b452](https://github.com/alliander-opensource/power-grid-model-io/pull/153/commits/da2b45246b5e3578bc5f9c240a01a3ed35bc47c9) in another barnch.

PR #160, opened 2023-02-24
use initialize array for dtypes instead of metadata

nitbharambe:


TonyXiang8787:
@nitbharambe we fix the issue in pgm in the latest version. This change is not needed.

PR #162, opened 2023-02-27
 Add data type arg in convert

nitbharambe:


bramstoeller (commented):
Technically you are breaking the interface of `covert()`, because you have added an extra argument in between other arguments. So please consider the following and then decide what you think fits best...

---
First of all, I don't think it is a practical issue in this case, because most people use keyword arguments but a call with non-keyword arguments, like below, will be broken:
```python
converter.convert(data, extra_info)
```
If people don't want to use keyword arguments, the new call is:
```python
converter.convert(data, None, extra_info)
```
The thing is that it may break silently at this point. `data_type` is not a string, it is the supplied extra_info dictionary and `extra_info` is None.

---

There are solutions to forbid non-keyword arguments, but I always feel that it's a bit against the python philosophy. An example would be:

```python
def fn(*, foo: int, bar: float):
   ...
```

So the proper way would be to add the new keyword argument last, or to make it non-optional. The later option is still breaking, but at least not silently.

```python
def convert(self, data: Dataset, extra_info: Optional[ExtraInfoLookup] = None, data_type: Optional[str] = None):
    ...

def convert(self, data: Dataset, data_type: str, extra_info: Optional[ExtraInfoLookup] = None):
    ...
```
---
A last, and maybe prettier suggestion (and useful in some other cases, like when you need to add extra arguments like `system_hz` or whatever) would be to just add and forward `**kwargs`:
```python
def convert(self, data: Dataset, extra_info: Optional[ExtraInfoLookup] = None, **kwargs):
   return self._serialize_data(data=data, extra_info=extra_info, **kwargs)
```

nitbharambe:
> 

Thanks for that detailed and clear explanation @bramstoeller ! 
I shall modify it in the way of the last suggestion with `**kwargs`

src/power_grid_model_io/converters/base_converter.py
```diff
@@ -92,7 +92,7 @@ def load_asym_output_data(self, data: Optional[T] = None) -> Dataset:
         data = self._load_data(data)
         return self._parse_data(data=data, data_type="asym_output", extra_info=None)
 
-    def convert(self, data: Dataset, extra_info: Optional[ExtraInfoLookup] = None) -> T:
+    def convert(self, data: Dataset, extra_info: Optional[ExtraInfoLookup] = None, **kwargs: Optional[str]) -> T:
```
bramstoeller, line 95: I think **kwargs can be for `Any` type. `Any` includes `None`, so it it per definition `Optional`.

src/power_grid_model_io/converters/base_converter.py
```diff
@@ -121,10 +125,13 @@ def save(
           extra_info: Optional[ExtraInfoLookup]:  (Default value = None)
           destination: Optional[BaseDataStore[T]]:  (Default value = None)
 
+        Keyword Args:
+          data_type: Optional[str]:   (Default value = None) Output data type "sym_output" or "asym_output"
```
bramstoeller, line 129: This is only true for the pandapower output. The keywords `sym_output` and `asym_output` are actually borrowed from power grid model, but you're using it to specify the pandapower output format. If woul dbe easy to check if a pgm attibute is sym (1 value) or asym (3 values), so you don't need an extra attribute for it.

bramstoeller: For example, the PgmJsonConverter doesn't care if the output data is sym or asym, it is written a way that it will convert pgm data to the corresponding json structure in any case.

nitbharambe:
Since we will be checking data type from the array itself, I shall close this PR and implement those 4 lines in the asym converter itself.

PR #163, opened 2023-02-28
Add trafo loading by current

nitbharambe:


bramstoeller (approved):
Funny, this looks good, but please consider my review for PR #162 as well.. We could do something similar (passing `**kwargs` to `_parse_data`) for these arguments too. But it's also justifiable that they belong to the constructor. I'll leave it up to you (and @petersalemink95) to decide what you feel is best.

nitbharambe:
> Funny, this looks good, but please consider my review for PR #162 as well.. We could do something similar (passing `**kwargs` to `_parse_data`) for these arguments too. But it's also justifiable that they belong to the constructor. I'll leave it up to you (and @petersalemink95) to decide what you feel is best.

Lets keep it this way for now. We need to add another parameter later when adding impedance component. This would need some thinking as to whether all these should be in constructor, `_parse_data` or `_serialize_data`. The `trafo_loading` is getting used only in output whereas `f_hz` and `base_mva` get used in input.

PR #164, opened 2023-02-28
Add pp input data to extra info

bramstoeller:


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -202,10 +237,28 @@ def _extra_info_to_pgm_input_data(self, extra_info: ExtraInfo):
...
             )
             for i, pgm_id in enumerate(data["id"]):
-                extra = extra_info[pgm_id]
+                extra = extra_info[pgm_id].get("pgm_input", {})
                 ref[i] = (pgm_id,) + tuple(extra[col] for col in node_cols)
             self.pgm_input_data[component] = ref
 
+    def _extra_info_to_pp_input_data(self, extra_info: ExtraInfo):
+        """
+        Converts extra component info into node_lookup
```
petersalemink95, line 246: Will other things than the derating factor (df) of transformers be stored here in the future? If so, then please mention in the docstring that only the derating factor (df) of transformers can be extracted from extra_info at the moment. 

If not, we should rename this function to something more meaningful, since this function would only focus on trafo df

petersalemink95 (approved):

petersalemink95 (approved):

PR #166, opened 2023-03-01
Lazy excel loading

bramstoeller:


tests/unit/data_stores/test_vision_excel_file_store.py
```diff
@@ -7,16 +7,16 @@
...
+def test_header_rows(mock_excel_file: MagicMock):
     # Arrange
     store = VisionExcelFileStore(file_path=Path("dummy.xlsx"))
-    read_excel_mock.return_value = {}
+    mock_excel_file.return_value.sheet_names = ["foo"]
 
     # Act
-    store.load()
+    data = store.load()
+    data["foo"]
```
nitbharambe, line 19: Kind of confused as to what this statement does.

bramstoeller: Yes, I can imagine that. Normally you would do something like:
```python
foo_data = data["foo"]
```
But in this case we are only testing the side effects. As this is a lazy loader, the side effect will be that the data is loaded, i.e. `parse()` is called, and specifically... it is called with the keyword argument `headers = [0, 1]`.

nitbharambe (approved):

PR #171, opened 2023-03-03
fix asym values validate in pp

nitbharambe:
Asymmetric values in input validation cannot not be validated because they are expected to be a singular value. Fixed that here.

### DIFF
--- a/tests/validation/utils.py
+++ b/tests/validation/utils.py
@@ -114,12 +114,15 @@ def select_values(actual: SingleDataset, expected: SingleDataset, component: str
     assert attribute in expected[component].dtype.names
 
     # Create an index series for both the actual data and the expected data
-    actual_values = pd.Series(actual[component][attribute], index=actual[component]["id"]).sort_index()
-    expected_values = pd.Series(expected[component][attribute], index=expected[component]["id"]).sort_index()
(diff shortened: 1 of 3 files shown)


tests/validation/utils.py
```diff
@@ -114,12 +114,15 @@ def select_values(actual: SingleDataset, expected: SingleDataset, component: str
     assert attribute in expected[component].dtype.names
 
     # Create an index series for both the actual data and the expected data
-    actual_values = pd.Series(actual[component][attribute], index=actual[component]["id"]).sort_index()
-    expected_values = pd.Series(expected[component][attribute], index=expected[component]["id"]).sort_index()
+    actual_attr = actual[component][attribute]
+    expected_attr = expected[component][attribute]
+    pd_data_fn = pd.DataFrame if actual_attr.ndim == expected_attr.ndim == 2 else pd.Series
```
petersalemink95, line 119: I'd do something like this:

`pd_data_fn = pd.DataFrame if actual_attr.ndim == 2 else pd.Series
assert actual_attr.ndim == expected_attr.ndim`

Why? If for some reason actual and expected do not match in dimension it would try to fit both in a Series and crash. With my proposal it would crash in the assert and the error would be self explanatory

PR #174, opened 2023-03-08
Fix Vision NL lookup

bramstoeller:


bramstoeller:
Extra info should match the Excel columns

PR #192, opened 2023-08-09
add missing API docs

mgovers:
* fix missing API documentation
* add panda power explanation of one-to-many ID map


mgovers:
c.c. @willemijnbrus

nitbharambe (approved):

PR #193, opened 2023-08-11
Add sc output function

nitbharambe:
The public function of load_sc_output_data is added.


tests/unit/converters/test_base_converter.py
```diff
@@ -117,6 +117,20 @@ def test_load_asym_output_data(converter: DummyConverter):
     assert data == {"foo": 1}
 
 
+def test_load_sc_output_data(converter: DummyConverter):
+    # Arrange
+    converter._parse_data.return_value = {"foo": 1}  # type: ignore
```
mgovers, line 122: Why do you need to pylint ignore?

tests/unit/converters/test_base_converter.py
```diff
@@ -117,6 +117,20 @@ def test_load_asym_output_data(converter: DummyConverter):
...
+
+    # Act
+    data = converter.load_sc_output_data(data={"node": [{"id": 1}, {"id": 2}]})
+
+    # Assert
+    converter._parse_data.assert_called_once_with(  # type: ignore
```
mgovers, line 128: Why do you need type ignore?

tests/unit/converters/test_base_converter.py
nitbharambe, line 122:
```suggestion
    converter._parse_data.return_value = {"foo": 1}
```

tests/unit/converters/test_base_converter.py
```diff
@@ -117,6 +117,20 @@ def test_load_asym_output_data(converter: DummyConverter):
...
 
+def test_load_sc_output_data(converter: DummyConverter):
+    # Arrange
+    converter._parse_data.return_value = {"foo": 1}
+
+    # Act
+    data = converter.load_sc_output_data(data={"node": [{"id": 1}, {"id": 2}]})
+
+    # Assert
+    converter._parse_data.assert_called_once_with(  # type: ignore
```
nitbharambe, line 128:
```suggestion
    converter._parse_data.assert_called_once_with( 
```

tests/unit/converters/test_base_converter.py
nitbharambe: @mgovers because those functions dont have those attributes. The same ignoring is done for other functions. I see some way of adding specs exist via mock but not sure if thats correct here. Do you know a better way?

mgovers: nah you can leave it be

mgovers:
@nitbharambe anything else that needs to be changed?

nitbharambe:
> @nitbharambe anything else that needs to be changed or can we merge?

Nothing more, can be merged!

mgovers (approved):

PR #196, opened 2023-10-04
Bump power grid model to 1.6.x

mgovers:


pyproject.toml
```diff
@@ -76,7 +76,7 @@ power_grid_model_io = ["config/**/*.yaml"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests/unit"]
-addopts = ["--cov=power_grid_model_io", "--cov-report=term", "--cov-report=html:cov_html", "--cov-fail-under=100"]
```
petersalemink95, line 79: Why do you want to decrease this? In normal circumstances the coverage is never allowed to decrease

mgovers: MyPY was not compliant (not even on `main`). I re-added it. because of that, I had to build in some redundant fail safety typechecks that can only be tested by testing private methods (which is generally something people are tempted to do when their design is poor)

100% coverage is actually virtually impossible when you do type checks, even for new code bases

tests/data/pandapower/pgm_asym_output_data.json
mgovers, line 1: The test data is **only** resorted and reformatted. I verified that by writing my own script that only does exactly that

tests/data/pandapower/pgm_input_data.json
mgovers, line 1: The test data is only resorted and reformatted. I verified that by writing my own script that only does exactly that

tests/data/pandapower/pgm_output_data.json
mgovers, line 1: The test data is only resorted and reformatted. I verified that by writing my own script that only does exactly that

.pre-commit-config.yaml
mgovers, line 1: these are not the latest versions but the versions also used by the PGM. Do you think it's a good idea to bump the versions to latest?

NOTE: the original combination of versions was broken, so this actually is necessary to fix #197

TonyXiang8787 (approved):

PR #202, opened 2023-10-10
Feature/more code sonar

mgovers:
* Fixes reliability sonar cloud issue
* (hopefully) fixes badge versions out of date 
image

### DIFF
--- a/src/power_grid_model_io/mappings/unit_mapping.py
+++ b/src/power_grid_model_io/mappings/unit_mapping.py
@@ -8,6 +8,7 @@ Unit mapping helper class
 from numbers import Number
 from typing import Dict, Optional, Set, Tuple
 
+import numpy as np
 import structlog
 
 #            si-unit   unit factor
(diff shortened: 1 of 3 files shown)


src/power_grid_model_io/mappings/unit_mapping.py
```diff
@@ -58,7 +59,7 @@ def set_mapping(self, mapping: Units):
                     )
                 self._mapping[unit] = (multiplier, si_unit)
                 if unit == si_unit:
-                    if multiplier != 1.0:
+                    if not np.isclose(multiplier, 1.0, rtol=1.0e-9, atol=0.0):
```
petersalemink95, line 62: Why is it needed to do it this way?

mgovers: sonar cloud tells us that for instance parsing strings `int("1.0")` may still give rounding errors, so comparing floats is dangerous

PR #204, opened 2023-10-24
Add download badges to README.md

mgovers:
cfr. https://github.com/PowerGridModel/power-grid-model/pull/410

Preview in https://github.com/PowerGridModel/power-grid-model-io/tree/feature/downloads-badge


TonyXiang8787:
The downloads of pgm-io is not impressive and this library is also not our main selling point. So my proposal is not to add it.

@mgovers you can add the download badge of pgm into our .github repo profile which will be shown in the organization home page.

mgovers:
> github

will do

PR #205, opened 2023-11-29
add substitution

nitbharambe:


src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -824,9 +824,11 @@ substitutions:
   ".*Switch state":
     "off": 0
     "in": 1
+    "on": 1
```
mgovers, line 827: please also update the tests/unit/mappings/test_value_mapping.py and the docs/converters/tabular_converter.md

mgovers: @nitish the documentation is still left: docs/converters/tabular_converter.md

mgovers: i would argue so because people will just blindly copy that into their own projects. we don't guarantee the docs to be up to date but they should be

mgovers (approved):

PR #213, opened 2024-01-10
Feature/dgc 2080 pandapower converter datatype different handling

Jerry-Jinfeng-Guo:
Mandatory 'expected_type' field to ensure pandapower input is proper

### DIFF
--- a/docs/examples/pandapower_example.ipynb
+++ b/docs/examples/pandapower_example.ipynb
@@ -84,7 +84,16 @@
    "metadata": {
     "scrolled": true
    },
-   "outputs": [],
+   "outputs": [
+    {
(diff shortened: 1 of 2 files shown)


mgovers:
Whoops apparently i clicked the wrong button here. Also, why was i able to merge this without approval? 🤔 we should fix this. @Jerry-Jinfeng-Guo do you know whether this pr actually broke things? If so, it should be reverted instead of just waiting for a follow-up pr

mgovers:
ohhh nvm i see you based #215 off the wrong branch and then reverted the commits instead of rebasing. that's why these commits are technically part of `main` and therefore this PR appears as `merged`. #217 is going to reapply them.

PR #219, opened 2024-01-30
DGC-2079 Logging level to base converter

Jerry-Jinfeng-Guo:
Initial implementation following @JeanDosSantos's input on Teams. Added log level, defaulted to ERROR but configurable. A small unit test is also included.

### DIFF
--- a/src/power_grid_model_io/converters/base_converter.py
+++ b/src/power_grid_model_io/converters/base_converter.py
@@ -7,7 +7,12 @@ Abstract converter class
 from abc import ABC, abstractmethod
 from typing import Generic, Optional, Tuple, TypeVar
 
+import logging
+logging.basicConfig(level=logging.ERROR)
+
 import structlog
+structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(logging.ERROR))
+
 from power_grid_model.data_types import Dataset, SingleDataset
 
 from power_grid_model_io.data_stores.base_data_store import BaseDataStore
@@ -20,10 +25,17 @@ T = TypeVar("T")
 class BaseConverter(Generic[T], ABC):
     """Abstract converter class"""
 
-    def __init__(self, source: Optional[BaseDataStore[T]] = None, destination: Optional[BaseDataStore[T]] = None):
+    def __init__(
+        self,
+        source: Optional[BaseDataStore[T]] = None,
+        destination: Optional[BaseDataStore[T]] = None,
+        log_level: int = logging.ERROR,
+    ):
         """
         Initialize a logger
         """
+        logging.basicConfig(level=log_level)
+        structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(log_level))
         self._log = structlog.get_logger(type(self).__name__)
         self._source = source
         self._destination = destination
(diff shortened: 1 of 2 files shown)


src/power_grid_model_io/converters/base_converter.py
```diff
@@ -7,7 +7,12 @@
 from abc import ABC, abstractmethod
 from typing import Generic, Optional, Tuple, TypeVar
 
+import logging
+logging.basicConfig(level=logging.ERROR)
+
 import structlog
+structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(logging.ERROR))
```
mgovers, line 14: python good practice: import before doing anything else, 2 whitelines after final import

src/power_grid_model_io/converters/base_converter.py
```diff
@@ -20,10 +25,17 @@
...
+        self,
+        source: Optional[BaseDataStore[T]] = None,
+        destination: Optional[BaseDataStore[T]] = None,
+        log_level: int = logging.ERROR,
+    ):
         """
         Initialize a logger
         """
+        logging.basicConfig(level=log_level)
+        structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(log_level))
```
mgovers, line 38: i don't know whether this is the case, but does this only override the local config within this class or the global config? if the latter, could it also affect users' config?

tests/unit/converters/test_base_converter.py
```diff
@@ -168,3 +168,13 @@ def test_load_data(converter: DummyConverter):
     converter_2 = DummyConverter(source=source)
     converter_2._load_data(data=None)
     source.load.assert_called_once()
+
+
+def test_base_converter_log_level():
+    import logging
```
mgovers, line 174: please move this import to the top. no harm in importing it for the other tests

src/power_grid_model_io/converters/base_converter.py
Jerry-Jinfeng-Guo: Just tried, not the global config rather than local instances: two logger script running at the same time could output their respective log level correctly

src/power_grid_model_io/converters/base_converter.py
Jerry-Jinfeng-Guo: Suggestions included in the newest commit.

src/power_grid_model_io/converters/base_converter.py
Jerry-Jinfeng-Guo: Conversation resolved.

tests/unit/converters/test_base_converter.py
Jerry-Jinfeng-Guo: Suggestion included in the newest commit. Conversation resolved.

Jerry-Jinfeng-Guo:
At the moment some of the tests are failing. They don't appear as directly related to the code added. In particular: `test_serialize_data` and others. Since pgm-io is still impossible for me to debug, could anyone take a look at the branch? Thanks!

mgovers:
> At the moment some of the tests are failing. They don't appear as directly related to the code added. In particular: `test_serialize_data` and others. Since pgm-io is still impossible for me to debug, could anyone take a look at the branch? Thanks!

I did some quick digging and it seems to have to do with the fact that the structlog configure is global

mgovers (approved):

PR #223, opened 2024-02-07
Feature/no global set option

mgovers:
* Fix almost all deprecation warnings, future warnings, etc.
* Do not suppress deprecation warnings in global config
* Warnings in pandapower in the validation tests are for some reason impossible to suppress

### DIFF
--- a/docs/examples/pgm_json_example.ipynb
+++ b/docs/examples/pgm_json_example.ipynb
@@ -498,7 +498,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.0"
+   "version": "3.10.5"
   }
  },
  "nbformat": 4,
--- a/docs/examples/vision_example.ipynb
+++ b/docs/examples/vision_example.ipynb
@@ -486,7 +486,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.0"
+   "version": "3.10.5"
   },
   "vscode": {
    "interpreter": {
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -2251,7 +2251,7 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         Returns:
             the "closed" value of a Switch
         """
-        switch_status = (
+        switch_states = (
             component[["index", bus]]
             .merge(
                 switches,
@@ -2263,7 +2263,7 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         )
 
         # no need to fill na because bool(NaN) == True
-        return pd.Series(switch_status.astype(bool, copy=False))
+        return pd.Series(switch_states.astype(bool, copy=False))
 
     def get_switch_states(self, pp_table: str) -> pd.DataFrame:
         """


tests/data/pandapower/pgm_output_data.json
mgovers, line 1: order of components changed

tests/data/pandapower/pgm_asym_output_data.json
mgovers, line 1: order of components changed

mgovers:
@Jerry-Jinfeng-Guo  there won't be a follow-up on the data format cause of the usage of the `PgmJsonConverter`, which we actually do not actively support anymore. @petersalemink95 needs to prioritize if we will ever update it to the new serialization format, and how. The main consideration there is the usage of the `extra_info`

tests/unit/converters/test_tabular_converter.py
```diff
@@ -980,9 +980,9 @@ def test_lookup_ids(converter: TabularConverter):
         reference,
         pd.DataFrame(
             [
-                ["foo", None, 1, 2, None],
-                ["node", None, 1, None, 2],
-                ["node", "bar", 1, 2, None],
+                ["foo", np.nan, 1, 2, np.nan],
+                ["node", np.nan, 1, np.nan, 2],
+                ["node", "bar", 1, 2, np.nan],
```
Jerry-Jinfeng-Guo, line 985: Is this change from `None` to `np.nan` due to change in pgm? Perhaps we could use another pair of eyes to confirm this because I don't have this knowledge without digging deep. (@petersalemink95 )

docs/examples/pandapower_example.ipynb
```diff
@@ -41,8 +41,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import warnings\n",
     "import pandapower as pp\n",
     "\n",
+    "warnings.filterwarnings(\"ignore\", module=\"pandapower\", category=FutureWarning) # Hide warnings related to pandas\n",
+    "\n",
```
Jerry-Jinfeng-Guo, line 48: This is indeed the better place to be.

tests/utils.py
```diff
@@ -172,21 +172,38 @@ def isnan(x: Any):
                 return False
 
         def eq(left, right) -> bool:
-            if isinstance(left, pd.DataFrame):
-                if left.columns != right.columns:
-                    return False
-            if isinstance(left, pd.Series):
-                if left.name != right.name:
-                    return False
+            if type(left) != type(right):
```
Jerry-Jinfeng-Guo, line 175: This `if` condition is new. Question without context: is this inferred from the old checks?

pyproject.toml
```diff
@@ -45,7 +45,8 @@ dev = [
     "pylint",
     "pytest",
     "pytest-cov",
-    "pydantic", # Used in unit tests
+    "pydantic>2", # Used in unit tests
+    "numba",
```
Jerry-Jinfeng-Guo, line 49: - I have the feeling that the test case is simple enough to justify the old `parse_obj_as` code?
- Are we doing any `jit` already?

Jerry-Jinfeng-Guo (commented):
Just a few questions

pyproject.toml
mgovers: 1. `parse_obj_as` is officially deprecated since Pydantic v2. With v3 likely coming next spring or early summer (see also https://docs.pydantic.dev/latest/version-policy/#pydantic-v3-and-beyond). since it just involves `dev` dependencies, living at HEAD is OK
2. `numba` was proposed by `pandapower` for the validation tests. it produced a warning and i saw no reason to keep the warning if this was all that was necessary to fix it and it's just a `dev` dependency

tests/utils.py
mgovers: short answer: yes.

long answer: `the truth value of an array is ambiguous` is quite hard to work around without stuff like this.

To make matters worse, potential `nan` values may be in the array, but may also be plain values. basically, if it is an array we need to check that all its components are either identical or nan both left and right. if it's anything else, if it is nan, then check that they both are. if not, they need to be equal. Just explaining this logic makes me want to puke but it is the way it is.

tests/unit/converters/test_tabular_converter.py
mgovers: no this is needed due to deprecated behaviour in `pandas` (according to `Mismatched null-like values nan and None found.`)

docs/examples/pgm_json_example.ipynb
```diff
@@ -498,7 +498,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.5"
+   "version": "3.12.0"
```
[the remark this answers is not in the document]
mgovers: fixed

docs/examples/vision_example.ipynb
```diff
@@ -486,7 +486,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.5"
+   "version": "3.12.0"
```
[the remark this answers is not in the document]
mgovers: fixed

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2248,18 +2251,19 @@ def get_individual_switch_states(component: pd.DataFrame, switches: pd.DataFrame
         Returns:
             the "closed" value of a Switch
         """
-        switch_state = (
+        switch_status = (
```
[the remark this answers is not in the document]
mgovers: fixed

Jerry-Jinfeng-Guo (approved):
We need to merge this PR asap. GOAT has already reported an error with old pandas.

Issue #225, opened 2024-02-14
[FEATURE] Support for multiple Vision Excel versions 


Jerry-Jinfeng-Guo:
Vision 9.7 has introduced many changes to the exported excel sheet, which brings challenge to the existing `VisionExcelConverter`. Most significantly, UUID was introduced as the new id scheme. However, older versions like <9.6 are still extensively used by users who wish to take advantage of PGM as their compute core. To address this, we need to update our code base to accomdate the changes.

Here is the proposed temporary work-around:
 - Process the new excel sheets in a separate routine and make it backwards compatible\
 - - The argument is simple: this approach requires minimum changes to the code base and PGM internally do not depend on UUID's, and there is hardly any benefits in its current applications.

In the case that `VisionExcelConverter` is to be continued, here are the proposed solutions:
- Open `*.yaml` configuration files as parameters to the converter. Currently one file is provided and users need to modify that versioned file or they need to go deep to the converter definition code to load any other configuration file. By making this a parameter, we could explicitly offload the whole `*.yaml` configuration task to users.
- Treat old `Number`/`Node.Number` the same as `GUID`/`Node.GUID` and convert them all to strings in python. They are then collected to a map where each string key maps to an integer. This way, we combines the existing id scheme well with both Vision 9.7 UUID based scheme and Vision sub 9.6 number based scheme.
- Extra care is needed trying to map terms: for instance, Vision 9.7 refer to `N1` now as `Grounding1`. Many other such changes need to be mapped out accurately.

mgovers:
Regardless of the path going forward, exposing the `*.yaml` configuration file is a good idea. Hasn't this been done already?

Jerry-Jinfeng-Guo:
Nope, not yet. Hence that will be the first PR in this issue.

mgovers:
> Nope, not yet. Hence that will be the first PR in this issue.

can you do that when you have time? i think this is important and small enough to "just do it"

Jerry-Jinfeng-Guo:
Step 1: https://github.com/PowerGridModel/power-grid-model-io/pull/226 @mgovers If you could give a review

Jerry-Jinfeng-Guo:
Step 0: https://github.com/PowerGridModel/power-grid-model-io/pull/227

Jerry-Jinfeng-Guo:
Step 1-3: https://github.com/PowerGridModel/power-grid-model-io/pull/232

mgovers:
[...]

PR #227, opened 2024-02-14
Temporary workaround converting Vision 97 file

Jerry-Jinfeng-Guo:
Standalone script as a temporaty workaround converting Vision 97 file to old id scheme. This relates to https://github.com/PowerGridModel/power-grid-model-io/issues/225


Jerry-Jinfeng-Guo:
This is a standalone script. The following two are to be decided:
- Not entirely sure where it should go. Currently it lives inside the utils directory. 
- Not sure if a test case is needed.

src/power_grid_model_io/utils/uuid_excel_cvtr.py
```diff
@@ -0,0 +1,176 @@
...
+    """Get the columns that contain the word "GUID"
+
+    Args:
+        df (pd.DataFrame): panda dataframe
+
+    Returns:
+        list: list of columns containing the word "GUID"
+    """
+    guid_regex = re.compile(r".*GUID")
+    return df.filter(regex=guid_regex).columns
```
mgovers, line 112: is there a guarantee that no accidental processions of the form `r".*GUID.*"` are matched?

if not, we need to be smarter about that.

if so, pandas documentation says it uses `re.search` behavior under the hood. In that case, just matching for `r"GUID"` should suffice combined with the explanation in the docstring

mgovers (commented):
most of it looks clear to me, but i don't particularly like the idea of having a temporary workaround script. what's the rationale?

src/power_grid_model_io/utils/uuid_excel_cvtr.py
Jerry-Jinfeng-Guo: Went with `(r".*GUID$")`

Jerry-Jinfeng-Guo:
> most of it looks clear to me, but i don't particularly like the idea of having a temporary workaround script. what's the rationale?

As mentioned in https://github.com/PowerGridModel/power-grid-model-io/issues/225#issue-2133779122, there are somethings to be decided within the team regarding the proper solutions and this temporary workaround serves as a compromise.

mgovers:
> This is a standalone script. The following two are to be decided:
> 
>     * Not entirely sure where it should go. Currently it lives inside the utils directory.

Please make a separate `scripts/` directory at the root.

>     * Not sure if a test case is needed. Currently code coverage will be 0.

I do think that's a good idea. A small one should suffice. It should also fix the overall coverage criterion on which CI currently fails

My expectation is that a semi-sustainable solution will involve including this code into the existing VSF converter. Actually, what is the reason you're not able to do this now?

Jerry-Jinfeng-Guo:
> > This is a standalone script. The following two are to be decided:
> > ```
> > * Not entirely sure where it should go. Currently it lives inside the utils directory.
> > ```
> 
> Please make a separate `scripts/` directory at the root.
> 
> > ```
> > * Not sure if a test case is needed. Currently code coverage will be 0.
> > ```
> 
> I do think that's a good idea. A small one should suffice. It should also fix the overall coverage criterion on which CI currently fails
> 
> My expectation is that a semi-sustainable solution will involve including this code into the existing VSF converter. Actually, what is the reason you're not able to do this now?

Like what is the reason for anything ;) The reason is that the proper solution (which I assume won't be simply converting the document) would take more time when this could be released to people to use already. I'll wait for some input on that one next Tuesday before I potentially spend more than necessary on it.

mgovers:
> Like what is the reason for anything ;) The reason is that the proper solution (which I assume won't be simply converting the document) would take more time when this could be released to people to use already. I'll wait for some input on that one next Tuesday before I potentially spend more than necessary on it.

that's why a small test case should suffice. it's useful to have in the first place

Jerry-Jinfeng-Guo:
> > Like what is the reason for anything ;) The reason is that the proper solution (which I assume won't be simply converting the document) would take more time when this could be released to people to use already. I'll wait for some input on that one next Tuesday before I potentially spend more than necessary on it.
> 
> that's why a small test case should suffice. it's useful to have in the first place but you don't have to spend too much time on it

Yup, agreed. It's on the way. But I have to leave for the appointment now. There was a tiny bug that was almost gone. New commit will be pushed and ready for review after lunch.

src/power_grid_model_io/scripts/__init__.py
mgovers, line 1: it's weird to have an `__init__.py` file in a `scripts/` directory. my bad, but see my comment below on the script itself

src/power_grid_model_io/scripts/uuid_excel_cvtr.py
mgovers, line 1: ohhhhh i see that i completely misunderstood what you meant with `script`. usually, `scripts` refer to something that can be called directly from the terminal.

since you actually have a class, `utils` was the correct location. Again: my bad

mgovers (approved):

PR #231, opened 2024-02-22
feature/log_output_doc

zhen0427:
pgm-io now provides an interface to configure the log level of the converter, we therefore add explanary information into the documentation.

### DIFF
--- a/docs/converters/converter.md
+++ b/docs/converters/converter.md
@@ -58,3 +58,23 @@ You can also add additional information about each component in the form of `ext
 ```python
 converter.save(example_data, extra_info=example_extra_info, destination=destination_path)
 ```
+
+## Configuring The Log Output
+ 
+We have provided an interface to configure the log level of the converter.
+Notice that this log level only belongs to the logger within the converter.
+Users need to set their basic configuration of the `logging` module to a level that is below what is configuired for the converters.
+ 
+```python
+# This is an example in your script
+import logging
+from power_grid_model_io.converters import VisionExcelConverter
+ 
+logging.basicConfig(level=logging.INFO) # Only levels INFO and above will be logged
+ 
+converter_warning = VisionExcelConverter(input_file, log_level=logging.WARNING) # If there is any logs above WARNING, they will be logged
+ 
+logging.basicConfig(level=logging.WARNING) # Only levels WARNING and above will be logged
+ 
+converter_debug = VisionExcelConverter(input_file, log_level=logging.DEBUG) # Any logs on DEBUG and INFO level will not be logged
+```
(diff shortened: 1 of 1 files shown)


docs/converters/converter.md
```diff
@@ -58,3 +58,23 @@ You can also add additional information about each component in the form of `ext
...
 ```
+
+## Configuring The Log Output
+ 
+We have provided an interface to configure the log level of the converter.
+Notice that this log level only belongs to the logger within the converter.
+Users need to set their basic configuration of the `logging` module to a level that is below what is configuired for the converters.
+ 
+```python
+# This is an example in your script
```
mgovers, line 69:
```suggestion
# An example of what may be included in your script
```

docs/converters/converter.md
```diff
@@ -58,3 +58,23 @@ You can also add additional information about each component in the form of `ext
...
+import logging
+from power_grid_model_io.converters import VisionExcelConverter
+ 
+logging.basicConfig(level=logging.INFO) # Only levels INFO and above will be logged
+ 
+converter_warning = VisionExcelConverter(input_file, log_level=logging.WARNING) # If there is any logs above WARNING, they will be logged
```
mgovers, line 75: it's probably good to mention that this does not affect the `logging.basicConfig` or any other converters created after that. maybe even provide an example of that, e.g.

```suggestion
converter_warning = VisionExcelConverter(input_file, log_level=logging.WARNING) # If there is any logs above WARNING, they will be logged
converter_info = VisionExcelConverter(input_file) # uses default INFO level
```

docs/converters/converter.md
```diff
@@ -58,3 +58,23 @@ You can also add additional information about each component in the form of `ext
...
+ 
+logging.basicConfig(level=logging.WARNING) # Only levels WARNING and above will be logged
```
mgovers, line 77: what happens if we change the `basicConfig` after a converter is created before this with default log level settings? e.g. does `converter_info` remain at level `INFO` or does it become `WARNING` at this declaration?

mgovers:
please update the description of the PR

Jerry-Jinfeng-Guo:
> please update the description of the PR

Indeed

docs/converters/converter.md
Jerry-Jinfeng-Guo: Reasonable rewording. Or something like:
```suggestion
# Examplary usage in your script
```

docs/converters/converter.md
Jerry-Jinfeng-Guo: I think the line below did just that?

docs/converters/converter.md
Jerry-Jinfeng-Guo: I agree with this comment and the one above.
```suggestion
logging.basicConfig(level=logging.WARNING) # Only levels WARNING and above will be logged
assert converter_info.get_log_level() == logging.INFO # previously created converters will retain their original log level, regardless of system wide configuration
```

Jerry-Jinfeng-Guo (approved):

PR #232, opened 2024-02-27
Feature/support UUID vision input

Jerry-Jinfeng-Guo:
Native support for UUID based id excel sheet from Vision >9.7 

Relates to #225

### DIFF
--- a/src/power_grid_model_io/converters/vision_excel_converter.py
+++ b/src/power_grid_model_io/converters/vision_excel_converter.py
@@ -37,17 +37,21 @@ class VisionExcelConverter(TabularConverter):
         self,
         source_file: Optional[Union[Path, str]] = None,
         language: str = "en",
-        terms_changed: dict = None,
+        terms_changed: Optional[dict] = None,
         mapping_file: Optional[Path] = None,
         log_level: int = logging.INFO,
-    ):
+    ):  # pylint: disable=too-many-arguments
         _mapping_file = _mapping_file = (
             mapping_file if mapping_file is not None else Path(str(DEFAULT_MAPPING_FILE).format(language=language))
         )
         if not _mapping_file.exists():
             raise FileNotFoundError(f"No Vision Excel mapping available for language '{language}'")
         self._id_reference: Optional[IdReferenceFields] = None
-        source = VisionExcelFileStore(file_path=Path(source_file), language=language, terms_changed=terms_changed) if source_file else None
+        source = (
+            VisionExcelFileStore(file_path=Path(source_file), language=language, terms_changed=terms_changed)
+            if source_file
+            else None
+        )
         super().__init__(mapping_file=_mapping_file, source=source, log_level=log_level)
 
     def set_mapping(self, mapping: Mapping[str, Any]) -> None:
--- a/src/power_grid_model_io/converters/vision_excel_converter.py
+++ b/src/power_grid_model_io/converters/vision_excel_converter.py
@@ -37,6 +37,7 @@ class VisionExcelConverter(TabularConverter):
         self,
         source_file: Optional[Union[Path, str]] = None,
         language: str = "en",
+        terms_changed: dict = None,
         mapping_file: Optional[Path] = None,
         log_level: int = logging.INFO,
     ):
@@ -46,7 +47,7 @@ class VisionExcelConverter(TabularConverter):
         if not _mapping_file.exists():
             raise FileNotFoundError(f"No Vision Excel mapping available for language '{language}'")
         self._id_reference: Optional[IdReferenceFields] = None
-        source = VisionExcelFileStore(file_path=Path(source_file), language=language) if source_file else None
+        source = VisionExcelFileStore(file_path=Path(source_file), language=language, terms_changed=terms_changed) if source_file else None
         super().__init__(mapping_file=_mapping_file, source=source, log_level=log_level)
 
     def set_mapping(self, mapping: Mapping[str, Any]) -> None:
--- a/src/power_grid_model_io/converters/vision_excel_converter.py
+++ b/src/power_grid_model_io/converters/vision_excel_converter.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from typing import Any, Mapping, Optional, Union
 
 from power_grid_model_io.converters.tabular_converter import TabularConverter
+from power_grid_model_io.data_stores.base_data_store import LANGUAGE_EN
 from power_grid_model_io.data_stores.vision_excel_file_store import VisionExcelFileStore
 
 DEFAULT_MAPPING_FILE = Path(__file__).parent.parent / "config" / "excel" / "vision_{language:s}.yaml"
@@ -36,13 +37,13 @@ class VisionExcelConverter(TabularConverter):
     def __init__(
         self,
         source_file: Optional[Union[Path, str]] = None,
-        language: str = "en",
+        language: str = LANGUAGE_EN,
         terms_changed: Optional[dict] = None,
-        mapping_file: Optional[Path] = None,
+        mapping_file: Optional[Union[Path, str]] = None,
         log_level: int = logging.INFO,
     ):  # pylint: disable=too-many-arguments
-        _mapping_file = _mapping_file = (
-            mapping_file if mapping_file is not None else Path(str(DEFAULT_MAPPING_FILE).format(language=language))
+        _mapping_file = Path(
+            mapping_file if mapping_file is not None else str(DEFAULT_MAPPING_FILE).format(language=language)
         )
         if not _mapping_file.exists():
             raise FileNotFoundError(f"No Vision Excel mapping available for language '{language}'")
--- a/src/power_grid_model_io/converters/vision_excel_converter.py
+++ b/src/power_grid_model_io/converters/vision_excel_converter.py
@@ -46,7 +46,7 @@ class VisionExcelConverter(TabularConverter):
         if not _mapping_file.exists():
             raise FileNotFoundError(f"No Vision Excel mapping available for language '{language}'")
         self._id_reference: Optional[IdReferenceFields] = None
-        source = VisionExcelFileStore(file_path=Path(source_file)) if source_file else None
+        source = VisionExcelFileStore(file_path=Path(source_file), language=language) if source_file else None
         super().__init__(mapping_file=_mapping_file, source=source, log_level=log_level)
 
     def set_mapping(self, mapping: Mapping[str, Any]) -> None:
--- a/src/power_grid_model_io/data_stores/excel_file_store.py
+++ b/src/power_grid_model_io/data_stores/excel_file_store.py
@@ -13,6 +13,7 @@ import pandas as pd
 
 from power_grid_model_io.data_stores.base_data_store import BaseDataStore
 from power_grid_model_io.data_types import LazyDataFrame, TabularData
+from power_grid_model_io.utils.uuid_excel_cvtr import UUID2IntCvtr, get_guid_columns, add_guid_values_to_cvtr, insert_or_update_number_column, update_column_names, special_nodes_en, special_nodes_nl
 
 
 class ExcelFileStore(BaseDataStore[TabularData]):
@@ -28,7 +29,7 @@ class ExcelFileStore(BaseDataStore[TabularData]):
 
     _unnamed_pattern: re.Pattern = re.compile(r"Unnamed: \d+_level_\d+")
 
-    def __init__(self, file_path: Optional[Path] = None, **extra_paths: Path):
+    def __init__(self, file_path: Optional[Path] = None, language: str = "en", **extra_paths: Path):
         super().__init__()
 
         # Create a dictionary of all supplied file paths:
@@ -45,6 +46,8 @@ class ExcelFileStore(BaseDataStore[TabularData]):
                 raise ValueError(f"{name} file should be a .xls or .xlsx file, {path.suffix} provided.")
 
         self._header_rows: List[int] = [0]
+        self._languange = language
+        self._uuid_cvtr = UUID2IntCvtr()
 
     def files(self) -> Dict[str, Path]:
         """
@@ -68,6 +71,7 @@ class ExcelFileStore(BaseDataStore[TabularData]):
                 sheet_data = xls_file.parse(xls_sheet_name, header=self._header_rows)
                 sheet_data = self._remove_unnamed_column_placeholders(data=sheet_data)
                 sheet_data = self._handle_duplicate_columns(data=sheet_data, sheet_name=xls_sheet_name)
+                sheet_data = self._process_uuid_columns(data=sheet_data, sheet_name=xls_sheet_name)
                 return sheet_data
 
             return sheet_loader
@@ -196,6 +200,29 @@ class ExcelFileStore(BaseDataStore[TabularData]):
                     to_rename[dup_idx] = f"{col_name[0]}_{counter}"
 
         return to_rename
+    
+    # def _process_uuid_columns(self, data: pd.DataFrame) -> pd.DataFrame:
+    def _process_uuid_columns(self, data: pd.DataFrame, sheet_name: str) -> pd.DataFrame:
+        first_level = data.columns.get_level_values(0)
+        guid_columns = first_level[first_level.str.endswith('GUID')]
+
+        for guid_column in guid_columns:
+            nr = 'Number' if self._languange == 'en' else 'Nummer'
+            add_guid_values_to_cvtr(data, guid_column, self._uuid_cvtr)
+            new_column_name = guid_column.replace("GUID", nr)
+            if guid_column == "GUID":
+                if sheet_name in special_nodes_en:
+                    new_column_name = guid_column.replace("GUID", "Subnumber")
+                elif sheet_name in special_nodes_nl:
+                    new_column_name = guid_column.replace("GUID", "Subnummer")
+            guid_column_pos = first_level.tolist().index(guid_column)
+            try:
+                data.insert(guid_column_pos + 1, new_column_name, data[guid_column].apply(self._uuid_cvtr.query))
+            except ValueError:
+                data[new_column_name] = df[guid_column].apply(self._uuid_cvtr.query)
+        
+        return data
+
 
     @staticmethod
     def _group_columns_by_index(data: pd.DataFrame) -> Dict[Union[str, Tuple[str, ...]], Set[int]]:
--- a/src/power_grid_model_io/data_stores/excel_file_store.py
+++ b/src/power_grid_model_io/data_stores/excel_file_store.py
@@ -11,9 +11,16 @@ from typing import Dict, List, Optional, Set, Tuple, Union
 
 import pandas as pd
 
-from power_grid_model_io.data_stores.base_data_store import BaseDataStore
+from power_grid_model_io.data_stores.base_data_store import (
+    DICT_KEY_NUMBER,
+    DICT_KEY_SUBNUMBER,
+    LANGUAGE_EN,
+    LANGUAGE_NL,
+    VISION_EXCEL_LAN_DICT,
+    BaseDataStore,
+)
 from power_grid_model_io.data_types import LazyDataFrame, TabularData
-from power_grid_model_io.utils.uuid_excel_cvtr import ( 
+from power_grid_model_io.utils.uuid_excel_cvtr import (
     UUID2IntCvtr,
     add_guid_values_to_cvtr,
     special_nodes_en,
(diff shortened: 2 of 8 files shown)


src/power_grid_model_io/converters/vision_excel_converter.py
```diff
@@ -37,16 +37,21 @@ def __init__(
         self,
         source_file: Optional[Union[Path, str]] = None,
         language: str = "en",
+        terms_changed: Optional[dict] = None,
         mapping_file: Optional[Path] = None,
```
mgovers, line 41: to keep the parallel as for the `source_file` argument: 

```suggestion
        mapping_file: Optional[Union[Path, str]] = None,
```

src/power_grid_model_io/converters/vision_excel_converter.py
```diff
@@ -37,16 +37,21 @@ def __init__(
...
         log_level: int = logging.INFO,
-    ):
+    ):  # pylint: disable=too-many-arguments
         _mapping_file = _mapping_file = (
             mapping_file if mapping_file is not None else Path(str(DEFAULT_MAPPING_FILE).format(language=language))
         )
```
mgovers, line 46: cfr. my above suggestion + redundant assignment
```suggestion
        _mapping_file = Path(
            mapping_file if mapping_file is not None else str(DEFAULT_MAPPING_FILE).format(language=language)
        )
```

src/power_grid_model_io/utils/uuid_excel_cvtr.py
```diff
@@ -159,6 +159,7 @@ def insert_or_update_number_column(
         number (str): "Number" or "Nummer" depending on the language
     """
     new_column_name = guid_column.replace("GUID", number)
+    # pylint: disable=duplicate-code
```
mgovers, line 162: cfr. my statement about the dictionary: both the dictionary itself and the sheet name language extraction may be moved to `utils` (which makes perfect sense). that would remove this `pylint` warning

tests/validation/converters/test_vision_excel_converter.py
```diff
@@ -297,3 +301,12 @@ def test_log_levels(capsys):
     cvtr5.set_log_level(logging.CRITICAL)
     outerr = capsys.readouterr()
     assert "debug" not in outerr.out
+
+
+def test_uuid_excel_input():
+    source_file = Path(str(SOURCE_FILE_97).format(language="en"))
+    ref_file_97 = convert_guid_vision_excel(source_file, number="Number", terms_changed=terms_chaged)
+    data_native, _ = VisionExcelConverter(source_file, language="en", terms_changed=terms_chaged).load_input_data()
```
mgovers, line 309: same here: please do a global search for `"en"` and `"nl"` and replace with `LANGUAGE_EN` and `LANGUAGE_NL` resp

src/power_grid_model_io/converters/vision_excel_converter.py
Jerry-Jinfeng-Guo: Unfortunately this will fail one of the legacy test

src/power_grid_model_io/converters/vision_excel_converter.py
```diff
@@ -37,16 +37,21 @@ def __init__(
         self,
         source_file: Optional[Union[Path, str]] = None,
         language: str = "en",
+        terms_changed: Optional[dict] = None,
         mapping_file: Optional[Path] = None,
         log_level: int = logging.INFO,
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: This argument came to me as well, but since they are parameters of different type, I went with the more logical order. Plus, positional parameters has its own share of problems and ideally should be discouraged in non-toy sccripts mo.

src/power_grid_model_io/converters/vision_excel_converter.py
mgovers: why? a `Union` type should not impact anything because the input type `Path` remains valid

src/power_grid_model_io/converters/vision_excel_converter.py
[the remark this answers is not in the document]
mgovers: 100% agree but not everyone is smart enough 😬

src/power_grid_model_io/converters/vision_excel_converter.py
Jerry-Jinfeng-Guo: This is not part of this PR but indeed could use the suggested rework

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -197,6 +216,31 @@ def _check_duplicate_values(self, sheet_name: str, data: pd.DataFrame) -> Dict[i
...
+
+        for guid_column in guid_columns:
+            nr = "Number" if self._languange == "en" else "Nummer"
+            add_guid_values_to_cvtr(data, guid_column, self._uuid_cvtr)
+            new_column_name = guid_column.replace("GUID", nr)
+            if guid_column == "GUID":
+                if sheet_name in special_nodes_en:
+                    new_column_name = guid_column.replace("GUID", "Subnumber")
+                elif sheet_name in special_nodes_nl:
+                    new_column_name = guid_column.replace("GUID", "Subnummer")
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Therefore I disabled pylint duplicated code warning

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -197,6 +216,31 @@ def _check_duplicate_values(self, sheet_name: str, data: pd.DataFrame) -> Dict[i
 
         return to_rename
 
+    def _process_uuid_columns(self, data: pd.DataFrame, sheet_name: str) -> pd.DataFrame:
+        first_level = data.columns.get_level_values(0)
+        guid_columns = first_level[first_level.str.endswith("GUID")]
+
+        for guid_column in guid_columns:
+            nr = "Number" if self._languange == "en" else "Nummer"
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: > for public libries, nl should not be the default.

The language was defaulted to 'en', not 'nl'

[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: The proposal was indeed the proper way, will include in the next commit.

tests/validation/converters/test_vision_excel_converter.py
Jerry-Jinfeng-Guo: I would confine this change only in this PR and not flood the whole code base. We should make another PR for it.

src/power_grid_model_io/data_stores/excel_file_store.py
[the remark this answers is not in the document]
mgovers: yea i understand that 😬 but with this rework, that duplicated code can actually be merged into one 😄

tests/validation/converters/test_vision_excel_converter.py
mgovers: ok, please do 😄

src/power_grid_model_io/data_stores/excel_file_store.py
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Yup. was my idea for the re-touching as well.

src/power_grid_model_io/converters/vision_excel_converter.py
Jerry-Jinfeng-Guo: I read the wrong document name xD

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -45,6 +66,9 @@ def __init__(self, file_path: Optional[Path] = None, **extra_paths: Path):
                 raise ValueError(f"{name} file should be a .xls or .xlsx file, {path.suffix} provided.")
 
         self._header_rows: List[int] = [0]
+        self._languange = language
```
mgovers, line 69:
```suggestion
        self._languange = language
        self._vision_excel_key_mapping = VISION_EXCEL_LAN_DICT[self._language]
```

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -197,6 +223,35 @@ def _check_duplicate_values(self, sheet_name: str, data: pd.DataFrame) -> Dict[i
...
+            new_column_name = guid_column.replace("GUID", nr)
+            if guid_column == "GUID":
+                if sheet_name in special_nodes_en:
+                    new_column_name = guid_column.replace(
+                        "GUID", VISION_EXCEL_LAN_DICT[LANGUAGE_EN][DICT_KEY_SUBNUMBER]
+                    )
+                elif sheet_name in special_nodes_nl:
+                    new_column_name = guid_column.replace(
+                        "GUID", VISION_EXCEL_LAN_DICT[LANGUAGE_NL][DICT_KEY_SUBNUMBER]
+                    )
```
mgovers, line 242: pull the mapping out of the for-loop or even in separate function.

```suggestion
        guid_columns = first_level[first_level.str.endswith("GUID")]
        sheet_key_mapping = {}
        if sheet_name in special_nodes_en:
            sheet_key_mapping = VISION_EXCEL_LAN_DICT[LANGUAGE_EN]
        elif sheet_name in special_nodes_nl:
            sheet_key_mapping = VISION_EXCEL_LAN_DICT[LANGUAGE_NL]

        for guid_column in guid_columns:
            nr = VISION_EXCEL_LAN_DICT[self._language][DICT_KEY_NUMBER]
            add_guid_values_to_cvtr(data, guid_column, self._uuid_cvtr)
            new_column_name = guid_column.replace("GUID", nr)
            if guid_column == "GUID":
                new_column_name = guid_column.replace(
                    "GUID", sheet_key_mapping[DICT_KEY_SUBNUMBER]
                )
```

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -197,6 +223,35 @@ def _check_duplicate_values(self, sheet_name: str, data: pd.DataFrame) -> Dict[i
...
+    def _process_uuid_columns(self, data: pd.DataFrame, sheet_name: str) -> pd.DataFrame:
+        first_level = data.columns.get_level_values(0)
+        guid_columns = first_level[first_level.str.endswith("GUID")]
+
+        for guid_column in guid_columns:
+            nr = VISION_EXCEL_LAN_DICT[self._languange][DICT_KEY_NUMBER]
+            add_guid_values_to_cvtr(data, guid_column, self._uuid_cvtr)
+            new_column_name = guid_column.replace("GUID", nr)
+            if guid_column == "GUID":
+                if sheet_name in special_nodes_en:
```
mgovers, line 235: is this criterion strict enough? what are the odds that there are duplicate values? If i look at the code, the mapping is quite arbitrary but maybe this is vision's own hack to make it work?

```python
special_nodes_en = [
    "Transformer loads",
    "Sources",
    "Synchronous generators",
    "Wind turbines",
    "Loads",
    "Capacitors",
    "Reactors",
    "Zigzag transformers",
    "Pvs",
]
special_nodes_nl = [
    "Transformatorbelastingen",
    "Netvoedingen",
    "Synchrone generatoren",
    "Windturbines",
    "Belastingen",
    "Condensatoren",
    "Spoelen",
    "Nulpuntstransformatoren",
    "Pv's",
]
```

on the same note: how likely is it that the sheet has a different language compared to the language provided to the excel file store?

src/power_grid_model_io/utils/uuid_excel_cvtr.py
```diff
@@ -161,9 +169,9 @@ def insert_or_update_number_column(
     new_column_name = guid_column.replace("GUID", number)
     if guid_column == "GUID":
         if sheet_name in special_nodes_en:
-            new_column_name = guid_column.replace("GUID", "Subnumber")
+            new_column_name = guid_column.replace("GUID", VISION_EXCEL_LAN_DICT[LANGUAGE_EN][DICT_KEY_SUBNUMBER])
         elif sheet_name in special_nodes_nl:
-            new_column_name = guid_column.replace("GUID", "Subnummer")
+            new_column_name = guid_column.replace("GUID", VISION_EXCEL_LAN_DICT[LANGUAGE_NL][DICT_KEY_SUBNUMBER])
```
mgovers, line 174: cfr before: can we remove this duplication taken into account my previous comments?

tests/validation/converters/test_vision_excel_converter.py
```diff
@@ -13,17 +13,24 @@
...
+from power_grid_model_io.utils.uuid_excel_cvtr import convert_guid_vision_excel
 
 from ..utils import compare_extra_info, component_attributes, component_objects, load_json_single_dataset, select_values
 
 DATA_PATH = Path(__file__).parents[2] / "data" / "vision"
 SOURCE_FILE = DATA_PATH / "vision_{language:s}.xlsx"
+SOURCE_FILE_97 = DATA_PATH / "vision_97_{language:s}.xlsx"
 VALIDATION_FILE = DATA_PATH / "pgm_input_data_{language:s}.json"
 LANGUAGES = ["en", "nl"]
+LANGUAGES_97 = ["en"]
```
mgovers, line 30: doesn't `nl` exist anymore? 😮

tests/validation/converters/test_vision_excel_converter.py
```diff
@@ -13,17 +13,24 @@
...
 VALIDATION_EN = Path(str(VALIDATION_FILE).format(language="en"))
 CUSTOM_MAPPING_FILE = DATA_PATH / "vision_9_5_{language:s}.yaml"
+terms_chaged = {"Grounding1": "N1", "Grounding2": "N2", "Grounding3": "N3", "Load.Behaviour": "Behaviour"}
```
mgovers, line 33:
```suggestion
terms_changed = {"Grounding1": "N1", "Grounding2": "N2", "Grounding3": "N3", "Load.Behaviour": "Behaviour"}
```

mgovers (commented):
i definitely like the rework. a couple remarks/questions/suggestions

src/power_grid_model_io/utils/uuid_excel_cvtr.py
Jerry-Jinfeng-Guo: The problem is that the `guid_column.replace(` only happens under the two conditions. I tried to initialize the `special_key_language` to `''` which corresponds to `.replace('GUID','GUID')`. It fails in some cases unfortunately and I figure to add the extra logic to prevent the `replace` from taking place, it would already be longer code than what is done currentlt.

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: Oops, should have seen it.

src/power_grid_model_io/utils/uuid_excel_cvtr.py
mgovers: but the other one also only loops over specifically `GUID` columns. if the logic becomes a separate function, the `== GUID` check may be either:

* pushed either outside the function (if the function itself does not have a `for` loop; or
* provided as an input for the function e.g. as `guid_columns=["GUID"]` (if the new function does contain a `for` loop over all `guid_columns`)

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: On this I would have to admit that I do not have enough knowledge. For now this is what I get from Vision 9.7. Not sure what would happen with the next release. Could use a follow-up to polish it if/when we know more.

tests/validation/converters/test_vision_excel_converter.py
Jerry-Jinfeng-Guo: I do not have a `nl` version test data yet

tests/validation/converters/test_vision_excel_converter.py
Jerry-Jinfeng-Guo: Oops.

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: Hahaha, I even had a typo there

src/power_grid_model_io/utils/uuid_excel_cvtr.py
Jerry-Jinfeng-Guo: Right. Solution comitted.

Jerry-Jinfeng-Guo:
Interesting. The last commit didn't trigger CI

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -28,9 +41,16 @@ class ExcelFileStore(BaseDataStore[TabularData]):
...
+    def __init__(
+        self,
+        file_path: Optional[Path] = None,
+        language: str = "en",
+        terms_changed: Optional[dict] = None,
+        **extra_paths: Path,
+    ):
         super().__init__()
-
+        if not isinstance(language, str):
```
mgovers, line 52: ? That's the users' responsibility tho. And we have mypy on our end

Or is this for compatibility? In that case we can also just bump mjnor version

tests/unit/utils/test_uuid_excel_cvtr.py
```diff
@@ -5,6 +5,9 @@
 from pathlib import Path
 
 from power_grid_model_io.converters import VisionExcelConverter
+from power_grid_model_io.data_stores.base_data_store import DICT_KEY_SUBNUMBER  # pylint: disable=unused-import
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: I put them there for a purpose ;)

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: Perhaps something for discussion next Tuesday.

Jerry-Jinfeng-Guo:
> One last consideration: should we maybe use an enum for the language? My guess is no but just calling it out so that we at least don't shoot ourselves in the foot later on

I don't think so either, strings are somewhat backward compatible and more intuitive. IMO `enum`s are nicer when there are multiple items but do not have a common way to group them. Language on the other hand... (https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#Table)

tests/unit/utils/test_uuid_excel_cvtr.py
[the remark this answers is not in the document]
mgovers: if they are to be added next PR, then please do so next PR and it's not necessary to add them here 😬

if it takes longer, it will probably just introduce confusion in the future without an explicit reason why you want to keep these import statements. We will forget about these things, or maybe some of the other team mates have a look at this code and will be like "why???"

tests/validation/converters/test_vision_excel_converter.py
```diff
@@ -13,17 +13,24 @@
 from power_grid_model.data_types import SingleDataset
 
 from power_grid_model_io.converters import VisionExcelConverter
+from power_grid_model_io.data_stores.base_data_store import DICT_KEY_SUBNUMBER  # pylint: disable=unused-import
+from power_grid_model_io.data_stores.base_data_store import LANGUAGE_NL  # pylint: disable=unused-import
```
mgovers, line 17:
```suggestion
```

same as for the `test_uuid_excel_cvtr.py`

src/power_grid_model_io/data_stores/excel_file_store.py
mgovers: Let's discuss during standup instead, then with a bit of luck we can decide on the path going forward and merge this PR today or tomorrow morning

mgovers (commented):
nitpicks we can discuss during standup at the latest. let's try to get this out the door. rework should take minutes at the most

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -28,9 +41,16 @@ class ExcelFileStore(BaseDataStore[TabularData]):
...
     _unnamed_pattern: re.Pattern = re.compile(r"Unnamed: \d+_level_\d+")
 
-    def __init__(self, file_path: Optional[Path] = None, **extra_paths: Path):
+    def __init__(
+        self,
+        file_path: Optional[Path] = None,
+        language: str = "en",
+        terms_changed: Optional[dict] = None,
+        **extra_paths: Path,
+    ):
```
Jerry-Jinfeng-Guo, line 50:
```suggestion
    def __init__(
        self,
        file_path: Optional[Path] = None,
        *,
        language: str = "en",
        terms_changed: Optional[dict] = None,
        **extra_paths: Path,
    ):
```

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: Included in the newest commmit.

src/power_grid_model_io/data_stores/excel_file_store.py
Jerry-Jinfeng-Guo: Included in the newest commmit.

mgovers (approved):

PR #240, opened 2024-04-04
Implement measurement field conversion from Vision

zhen0427:
only part of the code to determine the measured terminal type is finished. Further changes in the tabular converter is not implemented. Therefore pytest is still failed.

### DIFF
--- a/src/power_grid_model_io/functions/_functions.py
+++ b/src/power_grid_model_io/functions/_functions.py
@@ -11,7 +11,7 @@ from typing import Any, Optional, TypeVar, cast
 import numpy as np
 import structlog
 from power_grid_model import WindingType
-
+from power_grid_model import MeasuredTerminalType
 T = TypeVar("T")
 
 _LOG = structlog.get_logger(__file__)
@@ -24,13 +24,27 @@ WINDING_TYPES = {
     "ZN": WindingType.zigzag_n,
 }
 
+MEASURED_TERMINAL_TYPE_MAP = {
+    "transformer_from": MeasuredTerminalType.branch_from,
+    "transformer_to": MeasuredTerminalType.branch_to,
+    "cable_from": MeasuredTerminalType.branch_from,
+    "cable_to": MeasuredTerminalType.branch_to,
+    "transformer_load_to": MeasuredTerminalType.branch_to,
+    "load_to": MeasuredTerminalType.branch_to,
+    "earthing_transformer_to": MeasuredTerminalType.branch_to,
+    "transformer3_1": MeasuredTerminalType.branch3_1,
+    "transformer3_2": MeasuredTerminalType.branch3_2,
+    "transformer3_3": MeasuredTerminalType.branch3_3,
+}
+ 
+
 
 def has_value(value: Any) -> bool:
     """
     Return True if the value is not None, NaN or empty string.
     """
     if value is None:
-        return False
+        return False 
     if isinstance(value, float):
         return not np.isnan(value)
     return value != ""
@@ -104,3 +118,21 @@ def both_zeros_to_nan(value: float, other_value: float) -> float:
         _LOG.warning("0 replaced to nan")
         return float("nan")
     return value
+
+def find_terminal_type(**kwargs) -> MeasuredTerminalType:
+    """
+    Return 'branch_from' if any argument contains 'from' in its name.
+    Return 'branch_to' if any argument contains 'to' in its name.
+    Return the name of the first argument that evaluates to True otherwise.
+
+    Parameters:
+    - args: Variable number of input values.
+
+    Returns:
+    - String representation based on the conditions mentioned above.
+    """ 
+    for key, id in kwargs.items():
+        if id is not None:
+            return MEASURED_TERMINAL_TYPE_MAP[key]
+    _LOG.warning("No proper measured terminal type is found!")
+    return float("nan")
--- a/src/power_grid_model_io/functions/_functions.py
+++ b/src/power_grid_model_io/functions/_functions.py
@@ -25,16 +25,27 @@ WINDING_TYPES = {
 }
 
 MEASURED_TERMINAL_TYPE_MAP = {
-    "transformer_from": MeasuredTerminalType.branch_from,
-    "transformer_to": MeasuredTerminalType.branch_to,
     "cable_from": MeasuredTerminalType.branch_from,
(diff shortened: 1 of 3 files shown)


src/power_grid_model_io/utils/auto_id.py
```diff
@@ -119,4 +119,4 @@ def __getitem__(self, idx: int) -> Any:
         Returns:
             The original item
         """
-        return self._items[idx]
+        return getattr(self._items[idx], None)
```
Jerry-Jinfeng-Guo, line 122: What is the purpose of this change? pytest fails this one and passes when reverted to the original. (It should also be supplied with a `str` for the name field)

Jerry-Jinfeng-Guo (commented):
Not sure if ready for review, but since it's mentioned in the meeting today.

src/power_grid_model_io/functions/_functions.py
```diff
@@ -104,3 +128,13 @@ def both_zeros_to_nan(value: float, other_value: float) -> float:
...
     return value
+
+def find_terminal_type(**kwargs) -> MeasuredTerminalType:
+    """
+    Return the measured terminal type, based on the string representation
+    """ 
+    for key, id in kwargs.items():
+        if id is not None:
+            return MEASURED_TERMINAL_TYPE_MAP[key]
+    _LOG.warning("No measured terminal type is found!")
```
mgovers, line 139: what happens when `key` is not in `MEASURED_TERMINAL_TYPE_MAP`?

src/power_grid_model_io/utils/auto_id.py
mgovers: i indeed think that this is not what you intended to do. as it is now, besides the missing `str` type, it will basically try to seek the attribute `None` in `self._items[idx]`, e.g. `self._items[idx].None`

My guess is that you want to make it such that you return `None` if `idx` is not in `self_items`? in that case, you will want to use

```suggestion
        return self._items.get(idx)
```

which is shorthand for 

```python
        return self._items.get(idx, None)
```

(the `None` is the default argument)

nitbharambe:
@zhen0427 I suppose this function belongs to phase_to_phase functions
A test also needs to be in tests\unit\functions\test_phase_to_phase.py

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -242,26 +251,33 @@ def _convert_col_def_to_attribute(
         """
         # To avoid mistakes, the attributes in the mapping should exist. There is one extra attribute called
         # 'extra' in which extra information can be captured.
-        if attr not in pgm_data.dtype.names and attr != "extra":
+        if attr not in pgm_data.dtype.names and attr not in ["extra", "filter"]:
```
mgovers, line 254: pro tip: use `()` for inline things. it's one of the few cases where the python interpreter can optimize during parsing because tuples `()` are fixed-size. it's minor in this case, but this stuff accumulates multiplicatively so it can add up

```suggestion
        if attr not in pgm_data.dtype.names and attr not in ("extra", "filter"):
```

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -517,7 +533,7 @@ def _parse_auto_id(
             key_col_def: A column definition which should be unique for each object within the current table
 
         Returns: A single column containing numerical ids
-
+
```
mgovers, line 540: probably the formatter is already complaining so just calling it out

```suggestion

```

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -242,26 +251,33 @@ def _convert_col_def_to_attribute(
...
-        attr_data = self._parse_col_def(data=data, table=table, col_def=col_def, extra_info=extra_info)
+        attr_data = self._parse_col_def(data=data[table_mask], table=table, col_def=col_def, extra_info=extra_info)
 
         if len(attr_data.columns) != 1:
             raise ValueError(f"DataFrame for {component}.{attr} should contain a single column ({attr_data.columns})")
 
         pgm_data[attr] = attr_data.iloc[:, 0]
 
+    def _parse_filters() -> pd.Series[bool]:
+        mask = True * len(data) * 483
```
mgovers: * why specifically `483`? please name this (so-called) magic value
* i believe this doesn't do what you want because the current mask produces a scalar value, not a function or multi-dimensional array

src/power_grid_model_io/functions/_functions.py
```diff
@@ -129,12 +129,26 @@ def both_zeros_to_nan(value: float, other_value: float) -> float:
...
+    return float("nan")
+
+
+def if_not_link(input_string) -> bool:
+    """
+    Check if the measurement field is applied on a link
+    """
+    return input_string != "link"
+
+def filter_if_object(object_name: str, excl_object: str) -> bool:
```
mgovers: here and in other places: please consistently use 2 whitelines when there is no indentation, and 1 whiteline if there is indentation. e.g.since `def filter_if_object` is not indented, this should be prefixed with 2 whitelines
```suggestion
    return input_string != "link"


def filter_if_object(object_name: str, excl_object: str) -> bool:
```

tests/data/pandapower/pgm_asym_output_data.json
Jerry-Jinfeng-Guo, line 1: These two are output files and it's generally not necessary to be version them.

tests/data/pandapower/pgm_output_data.json
Jerry-Jinfeng-Guo, line 1: These two are output files and it's generally not necessary to be version them.

tests/unit/converters/test_tabular_converter.py
```diff
@@ -67,14 +67,14 @@ def test_set_mapping_file(converter: TabularConverter):
     with pytest.raises(ValueError, match="Mapping file should be a .yaml file, .txt provided."):
         converter.set_mapping_file(mapping_file=Path("dummy/path.txt"))
 
-    dummy_path = Path(__file__).parents[2] / "data" / "config" / "dummy_mapping.yaml"
+    dummy_path = Path(__file__).parents[2] / "data" / "config" / "dummy.yaml"
```
Jerry-Jinfeng-Guo, line 70: There is no `dummy.yaml` in the directory, instead there is a `mapping.yaml`

PR #256, opened 2024-06-14
filtering option in mapping

nitbharambe:
A filtering functionality can now be provided in the mapping. This is to not have an extra component for each row in excel if it doesnt exist.

- A filtering functionality in the mapping
- Note: The validation excel sheets are manually edited as vision 9.4/9.5 was not available to export the test data.


tests/data/vision/vision_validation.vnf
```diff
@@ -1,131 +1,143 @@
-V9.4
+V9.6
```
mgovers, line 1: shouldn't we have both vision versions next to eachother? we still have to keep `V9.4` alive, right?

nitbharambe: The difference is between 9.6 and 9.7.

nitbharambe:
Ready for preliminary review about the functionality. The conflicts with tests and different versions shall be resolved in parallel.

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -376,9 +407,15 @@ def _parse_col_def_const(data: TabularData, table: str, col_def: Union[int, floa
 
         """
         assert isinstance(col_def, (int, float))
-        return pd.DataFrame([col_def] * len(data[table]))
-
-    def _parse_col_def_column_name(self, data: TabularData, table: str, col_def: str) -> pd.DataFrame:
+        const_df = pd.DataFrame([col_def] * len(data[table]))
+        if table_mask is not None:
```
Jerry-Jinfeng-Guo, line 411: Not sure we need to double check if `np.sum(table_mask)>0` here

src/power_grid_model_io/functions/filters.py
```diff
@@ -0,0 +1,33 @@
...
+    filter out  by match value
+    """
+    return (row[col] != value).values[0]
+
+
+def filter_all_columns_empty_or_zero(row: pd.Series, cols: List[str]) -> bool:
+    """
+    filter out empty or zero values in multiple columns
+    """
+    return any(filter_by_value(row, col, 0) & filter_empty(row, col) for col in cols)
```
Jerry-Jinfeng-Guo, line 33: Would filtering out `zero` be desirable actually?

Jerry-Jinfeng-Guo (commented):
Overall question is:
- Not sure we need to double check if `np.sum(table_mask) > 0`
This applies to a couple places

src/power_grid_model_io/converters/tabular_converter.py
nitbharambe: This reminds me to check the =0 condition. Ill do so.

src/power_grid_model_io/functions/filters.py
nitbharambe: I guess it would be in the transformer loads context. In others, it would be about being empty.

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -178,24 +178,29 @@ def _convert_table_to_component(
...
         except KeyError as ex:
             raise KeyError(f"Invalid component type '{component}' or data type '{data_type}'") from ex
 
         if "id" not in attributes:
             raise KeyError(f"No mapping for the attribute 'id' for '{component}s'!")
 
         # Make sure that the "id" column is always parsed first (at least before "extra" is parsed)
-        sorted_attributes = sorted(attributes.items(), key=lambda x: "" if x[0] == "id" else x[0])
+        attributes_without_filter = {k: v for k, v in attributes.items() if k != "filter"}
+        sorted_attributes = sorted(attributes_without_filter.items(), key=lambda x: "" if x[0] == "id" else x[0])
```
mgovers, line 196: filter is usually about _including_ not _excluding_ (except it explicitly is mentioned as `exclude_filter` or just `exclude` or something similar). shouldn't this be the other way around to make it `filtered_attributes`?

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -204,12 +209,25 @@ def _convert_table_to_component(
...
+                table_mask &= data[table].apply(fn_ptr, axis=1, **kwargs).values
+        return table_mask
+
     # pylint: disable = too-many-arguments
     def _convert_col_def_to_attribute(
         self,
         data: TabularData,
         pgm_data: np.ndarray,
         table: str,
+        table_mask: np.ndarray,
```
mgovers, line 230: just `mask` is probably better than `table_mask`, no?

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -266,6 +291,7 @@ def _handle_extra_info(
         self,
         data: TabularData,
         table: str,
+        table_mask: np.ndarray,
```
mgovers, line 294: idem

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -353,17 +381,21 @@ def _parse_col_def(
...
+            )
         if isinstance(col_def, list):
-            return self._parse_col_def_composite(data=data, table=table, col_def=col_def)
+            return self._parse_col_def_composite(data=data, table=table, table_mask=table_mask, col_def=col_def)
         raise TypeError(f"Invalid column definition: {col_def}")
 
     @staticmethod
-    def _parse_col_def_const(data: TabularData, table: str, col_def: Union[int, float]) -> pd.DataFrame:
+    def _parse_col_def_const(
+        data: TabularData, table: str, col_def: Union[int, float], table_mask: Optional[np.ndarray] = None
```
mgovers, line 397: idem

tests/data/vision/pgm_input_data_en.json
```diff
@@ -6,49 +6,52 @@
...
-      {"id": 20, "node": 2, "id_reference": {"table": "Zigzag transformers", "key": {"Node.Number": 3, "Subnumber": 5}}, "Name": "zztrans1"},
-      {"id": 21, "node": 3, "id_reference": {"table": "Capacitors", "key": {"Node.Number": 4, "Subnumber": 9}}, "Name": "shunt1"},
-      {"id": 22, "node": 3, "id_reference": {"table": "Reactors", "key": {"Node.Number": 4, "Subnumber": 1}}}
+      {"id": 23, "node": 2, "id_reference": {"table": "Zigzag transformers", "key": {"Node.Number": 3, "Subnumber": 5}}, "Name": "zztrans1"},
+      {"id": 24, "node": 3, "id_reference": {"table": "Capacitors", "key": {"Node.Number": 4, "Subnumber": 9}}, "Name": "shunt1"},
+      {"id": 25, "node": 3, "id_reference": {"table": "Reactors", "key": {"Node.Number": 4, "Subnumber": 1}}}
+    ],
+  "three_winding_transformer":
+    [
+      {"id": 27, "node_1": 1, "node_2": 2, "node_3": 4, "id_reference": {"table": "Three winding transformers", "key": {"Number": 1}}}
```
mgovers, line 55: this is the crux, right? that this is filtered out. but now we do not have a test for 3w transformers anymore? that doesn't sound right either

tests/data/vision/vision_validation.vnf
mgovers, line 1: how do i review this?

tests/data/vision/vision_validation.vnf
```diff
@@ -1,131 +1,143 @@
-V9.4
+V9.7
```
mgovers, line 1: 😮

tests/data/vision/vision_validation.vnf
nitbharambe: Skip reviewing excel and vnf files I suppose. :D

tests/data/vision/pgm_input_data_en.json
```diff
@@ -6,49 +6,52 @@
...
+      {"id": 17, "node": 14, "id_reference": {"table": "Transformer loads", "name": "generation", "key": {"Node.Number": 3, "Subnumber": 2}}, "Name": "transformer_load_1"},
+      {"id": 18, "node": 14, "id_reference": {"table": "Transformer loads", "name": "pv_generation", "key": {"Node.Number": 3, "Subnumber": 2}}, "Name": "transformer_load_1"},
+      {"id": 20, "node": 3, "id_reference": {"table": "Synchronous generators", "key": {"Node.Number": 4, "Subnumber": 2}}, "Name": "syngen1"},
+      {"id": 21, "node": 2, "id_reference": {"table": "Wind turbines", "key": {"Node.Number": 3, "Subnumber": 3}}, "Name": "wind1"},
+      {"id": 26, "node": 2, "id_reference": {"table": "Pvs", "key": {"Node.Number": 3, "Subnumber": 1}}, "Name": "pv1"}
```
nitbharambe, line 41: @mgovers  The difference can be seen here, check the `Name`: transformer_load_2 is not visible in the sym_gens

tests/validation/converters/test_vision_excel_converter.py
```diff
@@ -332,11 +337,11 @@ def test_guid_extra_info():
     vision_cvtr = prep_vision_97(language=LANGUAGE_EN)
     _, extra_info = vision_cvtr.load_input_data()
 
-    assert extra_info[0]["GUID"] == "{8110C18E-7FDF-4A43-B868-E05E1DD7909F}"
-    assert extra_info[1]["GUID"] == "{D507B771-AC81-41F5-A292-91CF6119BE70}"
-    assert extra_info[2]["GUID"] == "{C92556BA-57E6-4780-A69E-DDDCA759BFBE}"
-    assert extra_info[3]["GUID"] == "{23C17CB9-7C1F-4AD0-B3C9-9B4AB99DB400}"
-    assert extra_info[4]["GUID"] == "{915748FF-2EE0-4C5E-806A-DBBFC433BEA4}"
+    assert extra_info[0]["GUID"] == "{7FF722ED-33B3-4761-84AC-A164310D3C86}"
```
nitbharambe, line 340: Should not have needed this. But somehow, a fresh export with this file made new GUIDs in the excel. Maybe because the vnf files version was changed. 
Maybe it originated because vision file from v9.4 to v9.7.

Something to keep in mind. @Jerry-Jinfeng-Guo

tests/data/vision/vision_validation.vnf
nitbharambe: Yes, If i open old 9.4 file, make modifications and save, it changes the version here too.

src/power_grid_model_io/converters/tabular_converter.py
nitbharambe: I like the idea about renaming this functionality to `exclude`. Ill do so.
This would then be named: `attributes_without_exclude_key`.

Jerry-Jinfeng-Guo: > filter is usually about including not excluding (except it explicitly is mentioned as exclude_filter or just exclude or something similar).

Hmm, not sure if I'd 100% agree there XD How would you understand terms like "filter out" then? I don't think there is a common sense of what filter would be including or excluding, it could even mean transforming. 

In our case, we are more thinking the purpose of sifting stuff like strainers, which is a perfect filtering process. Therefore, I would prefer Nitish's original wording of `attributes_without_filter `

src/power_grid_model_io/converters/tabular_converter.py
Jerry-Jinfeng-Guo: I think `table_mask` here is better because it carries the meaning of the fact that it's meant for `table`, instead of operations on `column` or `attribute` level

tests/validation/converters/test_vision_excel_converter.py
Jerry-Jinfeng-Guo: They are abusing UUID! We will run out of them sooner I'm afraid

tests/data/vision/vision_validation.vnf
Jerry-Jinfeng-Guo: @mgovers call Lauyrynas

mgovers: 😂

src/power_grid_model_io/converters/tabular_converter.py
mgovers: python has a built-in `filter` function that explicitly includes objects https://docs.python.org/3/library/functions.html#filter

the ambiguity is bad enough as is, but actually having name conflicts with python built-ins is even worse IMO

nitbharambe:
Unit test for filter pending. Converting to draft

tests/unit/converters/test_tabular_converter.py
mgovers, line 1: the amount of times that `np.ones` needs to be added for the `table_mask` in this file to me suggests that maybe it should have a default value

tests/unit/converters/test_tabular_converter.py
```diff
@@ -1017,3 +1165,40 @@ def test_lookup_ids__duplicate_keys(converter: TabularConverter):
...
+
+    mock_get_function.return_value = bool_fn_filter
+
+    actual = converter._parse_table_filters(data=tabular_data, table="nodes", filtering_functions=filtering_functions)
+
+    mock_get_function.assert_called_once_with("test_fn")
+    # check if return value is a 1d bool np array
+    assert isinstance(actual, np.ndarray)
+    assert actual.ndim == 1
+    assert actual.dtype == bool
```
mgovers, line 1196: can you actually compare dtypes using `==`? shouldn't that be `is`?

src/power_grid_model_io/config/excel/vision_en.yaml
```diff
@@ -424,6 +424,11 @@ grid:
       extra:
         - ID
         - Name
+      exclude_filter:
+        - power_grid_model_io.functions.exclude_filters.exclude_all_columns_empty_or_zero:
```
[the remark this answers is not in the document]
nitbharambe: It shall `exclude all objects where P and Q is empty or zero value`
What it evaluates to is irrelevant to the user. (For creating the mask, these exclude_filters should evaluate to False for a match)

Everything is included by default. So include filters is not needed.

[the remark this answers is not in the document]
mgovers: > It shall `exclude all objects where P and Q is empty or zero value`

I broke it down the other way around: It shall `"exclude": ["all objects where P and Q is empty or zero value"]`

> What it evaluates to is irrelevant to the user. (For creating the mask, these exclude_filters should evaluate to False for a match)

If so, I would just go with `filters` instead of `exclude_filter` and let it be an array of filters. Then, the filter `exclude_all_columns_empty_or_zero` would indeed do just that.

That's also more portable than adding just 1 filter option, because a user may want to add multiple `exclude`'s and maybe even `include_column_nonzero`. Otherwise, adding this feature would be a breaking change later on.

> Everything is included by default. So include filters is not needed.

That's only for now. I'm also thinking about the transformer tap regulator:

```yaml
"include_filters":
  - "checkbox_checked": "Transformer.regulated"
  - "checkbox_checked": "ThreeWindingTransformer.regulated"
```

or

```yaml
"filters":
  - "include_checkbox_checked": "Transformer.regulated"
  - "include_checkbox_checked": "ThreeWindingTransformer.regulated"
```

which reads more obviously than

```yaml
"exclude_filters":
  - "exclude_checkbox_unchecked": "Transformer.regulated"
  - "exclude_checkbox_unchecked": "ThreeWindingTransformer.regulated"
```

tests/unit/converters/test_tabular_converter.py
nitbharambe: Made it into default as None

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -242,20 +262,27 @@ def _convert_col_def_to_attribute(
         """
         # To avoid mistakes, the attributes in the mapping should exist. There is one extra attribute called
         # 'extra' in which extra information can be captured.
-        if attr not in pgm_data.dtype.names and attr != "extra":
+        if attr not in pgm_data.dtype.names and attr not in ["extra", "exclude_filter"]:
```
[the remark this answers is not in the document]
nitbharambe: Dont think we should have an include filter. All is included by default and then excludes from that

src/power_grid_model_io/config/excel/vision_en.yaml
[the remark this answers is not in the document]
nitbharambe: We don't have to complicate it. A single argument modification is applies to modification of 40 arguments in the code. Excel converter is temporary anyway. Its good to be least invasive.

mgovers (approved):

PR #264, opened 2024-07-08
Update doc on sheet name and duplicates

nitbharambe:


docs/converters/vision_converter.md
```diff
@@ -56,9 +56,12 @@ An examplery usage can be found in the example notebook as well as in the test c
...
 
 ### Different names for columns
-In different versions of exports, user can sometimes find different names for columns. For example, the column name `N1` might be represented as `Grounding1` in some exports. To address this, the `VisionExcelConverter`'s `terms_changed` argument can be used.
\ No newline at end of file
+In different versions of exports, user can sometimes find different names for columns. For example, the column name `N1` might be represented as `Grounding1` in some exports. To address this, the `VisionExcelConverter`'s `terms_changed` argument can be used.
+
+### Sheets being ignored
+In different versions of exports, user can find different names for the spreadsheets. All sheets not specified in the mapping file are ignored for conversion. Hence if the mapping has an incorrect sheet name, they would not be considered in conversion and no error shall be given out. The user should use the adequate mapping file for different vision versions. And if the default mapping is modified, it is advised to verify if all required sheets are being converted.
```
Jerry-Jinfeng-Guo, line 67:
```suggestion
In different versions of exports, user can find different names for the spreadsheets. All sheets not specified in the mapping file are ignored for conversion. Hence if the mapping has an incorrect sheet name, they would not be considered in conversion and no error shall be thrown. The user should use the adequate mapping file for different Vision versions. If the default mapping is modified, it is advised to verify if all required sheets are being converted.
```

docs/converters/vision_converter.md
```diff
@@ -56,9 +56,12 @@ An examplery usage can be found in the example notebook as well as in the test c
 So far we have the following issue known to us related to Vision exported spread sheets. We provide a solution from user perspective to the best of our knowledge.
 
 ### Duplicated `P` columns
-Vision can export sheets with duplicated `P` columns, one of which being unitless additional information. This field is of no actual purpose within PGM calculation. 
+Vision can export sheets with duplicated `P` columns, one of which being unitless additional information. This field is of no actual purpose within PGM calculation. In case of duplicate column names, PGM-IO renames them to eg. `P`, `P_2` and so on. However the change in mapping to the duplicate name should be done manually by the user in a custom mapping file.
```
Jerry-Jinfeng-Guo, line 59:
```suggestion
Vision can export sheets with duplicated `P` columns, one of which being unitless additional information. This field is of no actual purpose within PGM calculation. In case of duplicate column names, PGM-IO renames them e.g., from `P` to `P_2` and so on. However, the change in mapping to the duplicate name should be done manually by the user in a custom mapping file.
```

Jerry-Jinfeng-Guo (commented):
Almost there XD

docs/converters/vision_converter.md
nitbharambe: How about this?

```suggestion
Vision can export sheets with duplicated `P` columns, one of which being unitless additional information. This field is of no actual purpose within PGM calculation. In case of duplicate column names, PGM-IO renames them. The renaming is done in following way for example in case of a single duplicate:  `P` and `P_2`. However, the change in mapping to the duplicate name should be done manually by the user in a custom mapping file.
```

Jerry-Jinfeng-Guo (approved):

PR #266, opened 2024-07-25
Dont return empty array on filters

nitbharambe:


src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -180,6 +180,8 @@ def _convert_table_to_component(
 
         if "filters" in attributes:
             table_mask = self._parse_table_filters(data=data, table=table, filtering_functions=attributes["filters"])
+            if table_mask is not None and not table_mask.any():
+                return None
         else:
             table_mask = None
```
Jerry-Jinfeng-Guo, line 187: In both branches, `table_mask` can be `None` (the first one from the `_parse_table_filters`). Would that behavior still be desirable?

nitbharambe: The returning value for this function: `_convert_table_to_component` as `None` ignores the component and does not create any array.

The `table_mask=None` for all functions called inside this do different thing. It "ignores" the mask. ie. equivalent to all are True. The default of `None` for table_mask was selected to not make changes to the tests everywhere. The table_mask would have gotten passed at almost all functions otherwise.

Jerry-Jinfeng-Guo: So when `data[table]` yeilds a simple list other than a panda data frame, filters won't apply here? In that case I wonder whether it is necessary to do the first check inside the function `_parse_table_filters`

nitbharambe: Yes, when it yields a np.ndarray  specifically, filters wont be applicable. 

So is the case for all of functions within mapping for example. They all are used with `.apply` which do not work with ndarray while TabularData can be composed of ndarray.

Jerry-Jinfeng-Guo (approved):
LGTM

PR #270, opened 2024-08-21
Black jupyter precommit

nitbharambe:
Black can format jupyter notebooks too. Hence added it to pre-commit


nitbharambe:
> Should we also add it to the pyproject toml `[docs]` optional dependencies? If not, that's also fine with me

I dont think so. Its not a "dependency" as such.

PR #275, opened 2024-09-30
Feature/modernize ci

mgovers:
fixes CI

This blocks merging of #274


src/power_grid_model_io/converters/pandapower_converter.py
mgovers, line 1: all these changes were caused by version bumps in pylint and black. It may be a good idea to include this in the Nightly build

mgovers: see #276

Jerry-Jinfeng-Guo (approved):

PR #279, opened 2024-10-20
Update arrow example for columnar data

nitbharambe:

- Added columnar data example (WIP based on decisions)
- Schema creation

- Table creation is changed to record batch

### Decisions

- Keep or remove row based examples? 
  - Decided to only keep columnar
- Table creation should be individually or via the schema?  Check `#TODO Decisions: Create from schema`
- Asymmetric attributes should be fixed list array or individual phases or both?  Check `# TODO Decisions: Asymmetric attributes should be fixed list array or individual phases or both?`
- Should we leave update data creation upto users again or provide an example for that as well?


docs/examples/arrow_example.ipynb
```diff
@@ -92,6 +107,166 @@
...
+    }
+   ],
+   "source": [
+    "#TODO Decisions: Create from schema\n",
+    "pa.record_batch(\n",
+    "    {\n",
+    "        \"id\": [1, 2, 3],\n",
+    "        \"u_rated\": [10500.0, 10500.0, 10500.0],\n",
+    "    },\n",
+    "    schema=pgm_schema(DatasetType.input, ComponentType.node),\n",
```
mgovers, line 266: > Table creation should be individually or via the schema?

I don't fully understand what you mean but I suppose you mean whether to use `power_grid_meta_data`? if so: yes, please do use that to deduce the data types

docs/examples/arrow_example.ipynb
```diff
@@ -583,11 +841,227 @@
...
+       "q_specified: [[0.5,1500,0.1],[1.5,2.5,1500]]"
+      ]
+     },
+     "execution_count": 27,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# TODO Decisions: Asymmetric attributes should be fixed list array or individual phases or both?\n",
```
mgovers, line 969: since we only support 3phase asym attributes, let's keep it as is. the 3-dimensional struct for asymmetric is - i believe - the best we can do. or what exactly do you mean?

mgovers (commented):
> Should we leave update data creation up to users again or provide an example for that as well?

since this is only an example notebook, not a full feature, i don't think we need to add that. the only difference would be `DatasetType.update` instead of `DatasetType.input`

sidenote: i see you add a lot of new cells. is there a need for that? or is it because you haven't removed the old ones yet?

docs/examples/arrow_example.ipynb
nitbharambe: Yes, table creation by power_grid_meta_data instead of manually specifying types.
Ill change to use from meta.

docs/examples/arrow_example.ipynb
nitbharambe: From all conversations I can definitely confirm that having zero copy conversion is the main goal here. It was not possible via "p_specified_a", "p_specified_b", "p_specified_c". A Fixed list array can do it with few operations. Hence I shall go ahead and modify that too. 

It shall be clear when the PR is in review.

mgovers: Arrow doesn't really support multi-dimensional columns (AFAIK) so zero-copy can't be avoided here, most likely. you can investigate whether it is possible, but if not, don't spend too much time on it. other people with more knowledge of Arrow may optimize this.

nitbharambe:
> > Should we leave update data creation up to users again or provide an example for that as well?
> 
> since this is only an example notebook, not a full feature, i don't think we need to add that. the only difference would be `DatasetType.update` instead of `DatasetType.input`
> 
> sidenote: i see you add a lot of new cells. is there a need for that? or is it because you haven't removed the old ones yet?

Cells regarding decisions will be removed. The final example will be similar size as original.

docs/examples/arrow_example.ipynb
nitbharambe: I changed the wording. I am not sure how to mention "guaranteed" when an advanced user can indeed guarantee it if custom types are built correctly. :D

docs/examples/arrow_example.ipynb
```diff
@@ -92,6 +118,73 @@
...
+    "            continue\n",
+    "        if dtype.shape == (3,):\n",
+    "            pa_dtype = pa.list_(pa.from_numpy_dtype(dtype.base), 3)\n",
+    "        else:\n",
+    "            pa_dtype = pa.from_numpy_dtype(dtype)\n",
+    "        schemas.append((meta_attribute, pa_dtype))\n",
+    "    return pa.schema(schemas)\n",
+    "\n",
+    "\n",
+    "print(\"-------node combined asym scehma-------\")\n",
```
mgovers, line 182: phrasing is a bit weird. maybe something like `node pyarrow data schema`?

docs/examples/arrow_example.ipynb
```diff
@@ -105,88 +198,65 @@
...
+    "sources_dict = {\"id\": [6], \"node\": [1], \"status\": [1], \"u_ref\": [1.0]}\n",
+    "\n",
+    "sym_loads_dict = {\n",
+    "    \"id\": [7, 8],\n",
+    "    \"node\": [2, 3],\n",
+    "    \"status\": [1, 1],\n",
+    "    \"type\": [0, 0],\n",
+    "    \"p_specified\": [1.0, 2.0],\n",
+    "    \"q_specified\": [0.5, 1.5],\n",
+    "}\n",
```
mgovers, line 249: since most of the people diving into this example notebook are already using pyarrow (or get it directly from the database), they most likely have `pa.table` or at least pyarrow-native data types. i would prefer if you can create a full table/record batch directly in pyarrow (like it was before). For the array, you may be able to use `from_arrays` (https://arrow.apache.org/docs/python/generated/pyarrow.FixedSizeListArray.html#pyarrow.FixedSizeListArray.from_arrays ) or even create a separate table just for the sake the asymmetric value.

alternatively, we can just say that "there's currently no simple zero-copy solution for asymmetric data but here's a solution with copy"

docs/examples/arrow_example.ipynb
```diff
@@ -195,42 +265,52 @@
...
     "    \"\"\"Convert Arrow data to NumPy data.\"\"\"\n",
-    "    result = initialize_array(data_type, component, len(data))\n",
+    "    result = {}\n",
+    "    result_dtype = power_grid_meta_data[dataset_type][component_type].dtype\n",
     "    for name, column in zip(data.column_names, data.columns):\n",
-    "        if name in result.dtype.names:\n",
-    "            result[name] = column.to_numpy()\n",
+    "        column_data = column.to_numpy(zero_copy_only=zero_copy_only)\n",
+    "        if zero_copy_only and column_data.dtype != result_dtype[name]:\n",
+    "            raise ValueError(ZERO_COPY_ERROR_MSG)\n",
```
mgovers, line 305: or just
```suggestion
    "        assert column_data.dtype == result_dtype[name]\n",
```
then you don't need the note or the argument. it's just for illustration purposes anyway

docs/examples/arrow_example.ipynb
```diff
@@ -195,42 +265,52 @@
    "source": [
     "### Convert the Arrow data to power-grid-model input data\n",
     "\n",
-    "No direct conversion from Arrow Tables to NumPy exists and a copy is always required.\n",
+    "The Arrow record batch or tables can then be converted to row based data or columnar data.\n",
+    "Converting Arrow data to columnar NumPy arrays is recommended to leverage the columnar nature of Arrow data. \n",
+    "This conversion can be done with zero-copy operations.\n",
+    "\n",
+    "Similar approach be adopted by the user to convert to row based data.\n",
```
mgovers, line 272:
```suggestion
    "A similar approach be adopted by the user to convert to row based data.\n",
```

mgovers:
CI is failing on code quality

docs/examples/arrow_example.ipynb
```diff
@@ -77,7 +103,7 @@
       "node: {'names': ['id', 'u_rated'], 'formats': ['<i4', '<f8'], 'offsets': [0, 8], 'itemsize': 16, 'aligned': True}\n",
       "line: {'names': ['id', 'from_node', 'to_node', 'from_status', 'to_status', 'r1', 'x1', 'c1', 'tan1', 'r0', 'x0', 'c0', 'tan0', 'i_n'], 'formats': ['<i4', '<i4', '<i4', 'i1', 'i1', '<f8', '<f8', '<f8', '<f8', '<f8', '<f8', '<f8', '<f8', '<f8'], 'offsets': [0, 4, 8, 12, 13, 16, 24, 32, 40, 48, 56, 64, 72, 80], 'itemsize': 88, 'aligned': True}\n",
       "source: {'names': ['id', 'node', 'status', 'u_ref', 'u_ref_angle', 'sk', 'rx_ratio', 'z01_ratio'], 'formats': ['<i4', '<i4', 'i1', '<f8', '<f8', '<f8', '<f8', '<f8'], 'offsets': [0, 4, 8, 16, 24, 32, 40, 48], 'itemsize': 56, 'aligned': True}\n",
-      "asym_load: {'names': ['id', 'node', 'status', 'type', 'p_specified', 'q_specified'], 'formats': ['<i4', '<i4', 'i1', 'i1', ('<f8', (3,)), ('<f8', (3,))], 'offsets': [0, 4, 8, 12, 16, 40], 'itemsize': 64, 'aligned': True}\n"
+      "asym_load: {'names': ['id', 'node', 'status', 'type', 'p_specified', 'q_specified'], 'formats': ['<i4', '<i4', 'i1', 'i1', ('<f8', (3,)), ('<f8', (3,))], 'offsets': [0, 4, 8, 9, 16, 40], 'itemsize': 64, 'aligned': True}\n"
```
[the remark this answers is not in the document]
nitbharambe: Wait, how can this diff of offsets be explained? :/

docs/examples/arrow_example.ipynb
nitbharambe: Sure is, forgot to modify it.

docs/examples/arrow_example.ipynb
nitbharambe: I made this change to make it cleaner. Mentioning individual types is a bit messy.
It is upto the user how they wish to create the array. In the end, it should match the schema created in the helper function which is created from power_grid_meta_data.

Its similar reasoning as to how we have power flow example as
```
lines = initialize_array()
lines["id"] = [1,2,3]
```
instead of 
```
lines = initialize_array()
lines["id"] = np.array([1,2,3], dtype=np.int32())
```

mgovers: the danger with record batch is that it's a view of the data, not actual pyarrow data. it can be anything: contiguous, strided, fortran order, ... The whole point of this notebook is to show actual arrow data being loaded into the PGM.

docs/examples/arrow_example.ipynb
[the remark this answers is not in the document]
mgovers: compiler version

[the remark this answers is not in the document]
nitbharambe: Oh I remember, offsets are determined by numpy automatically.

docs/examples/arrow_example.ipynb
```diff
@@ -212,47 +207,60 @@
...
     "\n",
-    "sym_loads_dict = {\n",
-    "    \"id\": [7, 8],\n",
-    "    \"node\": [2, 3],\n",
-    "    \"status\": [1, 1],\n",
-    "    \"type\": [0, 0],\n",
-    "    \"p_specified\": [1.0, 2.0],\n",
-    "    \"q_specified\": [0.5, 1.5],\n",
-    "}\n",
+    "lines = pa.record_batch(\n",
```
mgovers, line 225:
```suggestion
    "# or convert directly using the schema\n"
    "lines = pa.record_batch(\n",
```

docs/examples/arrow_example.ipynb
```diff
@@ -212,47 +207,60 @@
...
-    "    \"c1\": [4.1e-05, 5.4e-05],\n",
-    "    \"tan1\": [0.1, 0.1],\n",
-    "    \"r0\": [0.01, 0.05],\n",
-    "    \"x0\": [0.22, 0.06],\n",
-    "    \"c0\": [4.1e-05, 5.4e-05],\n",
-    "    \"tan0\": [0.4, 0.1],\n",
-    "}\n",
-    "\n",
-    "sources_dict = {\"id\": [6], \"node\": [1], \"status\": [1], \"u_ref\": [1.0]}\n",
+    "nodes_schema = pgm_schema(DatasetType.input, ComponentType.node)\n",
```
mgovers, line 216:
```suggestion
    "# create the individual columns with the correct data type\n",
    "nodes_schema = pgm_schema(DatasetType.input, ComponentType.node)\n",
```

mgovers: these were not included in the latest run of readthedocs, it would appear. is that correct?

docs/examples/arrow_example.ipynb
```diff
@@ -195,42 +269,50 @@
...
+    "Converting Arrow data to columnar NumPy arrays is recommended to leverage the columnar nature of Arrow data. \n",
+    "This conversion can be done with zero-copy operations.\n",
+    "\n",
+    "A similar approach be adopted by the user to convert to row based data.\n",
     "\n",
-    "To ensure support for optional arguments and to prevent version lock, it is recommended to create an empty power-grid-model data set using `initialize_array` and then fill it with the Arrow data."
+    "```{note}\n",
+    "The option of `zero_copy_only` in the function below and assert for correct dtype is added in this demo to verify no copies are made. \n",
+    "Its usage is not mandatory to do zero copy conversion.\n",
+    "```"
```
mgovers, line 281: this note is outdated since you removed the `zero_copy_only`. either update and/or add a comment in the function below in the code

mgovers (commented):
most of it looks good

docs/examples/arrow_example.ipynb
nitbharambe:
```suggestion
    "```{note}\n",
    "The option of `zero_copy_only` in the `to_numpy` function below and assert for correct dtype is added in this demo to verify no copies are made. \n",
    "Its usage is not mandatory to do zero copy conversion.\n",
    "```"
```
I had it added also because unlike numpy's `copy=False`, pyarrow's `zero_copy_only=True` is a hard check which raises error. So an extra note that such hard check is not mandatory.

mgovers: just an inline comment will do then

docs/examples/arrow_example.ipynb
```diff
@@ -195,42 +270,47 @@
    "source": [
     "### Convert the Arrow data to power-grid-model input data\n",
     "\n",
-    "No direct conversion from Arrow Tables to NumPy exists and a copy is always required.\n",
+    "The Arrow record batch or tables can then be converted to row based data or columnar data.\n",
+    "Converting Arrow data to columnar NumPy arrays is recommended to leverage the columnar nature of Arrow data. \n",
+    "This conversion can be done with zero-copy operations.\n",
     "\n",
-    "To ensure support for optional arguments and to prevent version lock, it is recommended to create an empty power-grid-model data set using `initialize_array` and then fill it with the Arrow data."
+    "A similar approach be adopted by the user to convert to row based data."
```
figueroa1395, line 277:
```suggestion
    "A similar approach can be adopted by the user to convert to row based data instead."
```

docs/examples/arrow_example.ipynb
```diff
@@ -195,42 +270,47 @@
    "source": [
     "### Convert the Arrow data to power-grid-model input data\n",
     "\n",
-    "No direct conversion from Arrow Tables to NumPy exists and a copy is always required.\n",
+    "The Arrow record batch or tables can then be converted to row based data or columnar data.\n",
```
figueroa1395, line 273: "The Arrow record batch or tables" sounds weird, but once one realices it is refering to `record_batch` or `tables`, it is more clear. So is it possible to enclose those in `` or something similar to indicate code?

docs/examples/arrow_example.ipynb
```diff
@@ -295,12 +354,14 @@
    "source": [
     "### Use the power-grid-model\n",
     "\n",
-    "For more extensive examples, visit the [power-grid-model documentation](https://power-grid-model.readthedocs.io/en/stable/index.html)."
+    "The `output_component_types` argument is set to `ComponentAttributeFilterOptions.relevant` to given out columnar data.\n",
```
figueroa1395, line 357: I think that the other comment was about making sure the output is always columnar to avoid again additional copies.

figueroa1395 (commented):
Additional nitpick: Although the section for batch data was not updated, I would still add a sentence reminding that it should be provided in columnar format to avoid zero copy (right?) and that it can also be validated. Just to be more self contained, but it is fine as is if you don't deem it necessary.

nitbharambe:
> Additional nitpick: Although the section for batch data was not updated, I would still add a sentence reminding that it should be provided in columnar format to avoid zero copy (right?) and that it can also be validated. Just to be more self contained, but it is fine as is if you don't deem it necessary.

A lot more details can be mentioned about  batch data, hence I chose not to mention any of them as it might appear incomplete :D
This point about columnar can indeed be inferred by the user.

In batch case, there would be multiple things to think about: 
Different data sources, different batches with multiple scenarios etc. An advanced user would definitely needs to dive deeper there.

figueroa1395 (approved):

docs/examples/arrow_example.ipynb
```diff
@@ -142,17 +136,17 @@
...
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "-------node combined asym scehma-------\n",
+      "-------node scehma-------\n",
       "id: int32\n",
       "u_rated: double\n",
-      "-------asym load combined asym scehma-------\n",
+      "-------asym load scehma-------\n",
```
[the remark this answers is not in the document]
mgovers: how did this end up in production? please do a follow-up PR to fix this. also double-check that there are no other typos like this remaining

PR #294, opened 2025-04-04
Restrict pandapower to <3.0

nitbharambe:
CI fails at #292 and #293 . We need to 
1) upgrade pandapower to 3.0 
2) remove the `__array_prepare__` from MockFn before upgrade to numpy 2.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers=[
 ]
 requires-python = ">=3.11"
 dependencies = [
-    "numpy>=1.20, <2.0",
+    "numpy>=1.20",
     "openpyxl",
     "pandas",
     "power_grid_model>=1.8",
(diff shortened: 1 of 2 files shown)


mgovers:
I did a quick experiment and this can be safely done without any impact on the repo:

> remove the __array_prepare__ from MockFn before upgrade to numpy 2.0

nitbharambe:
> I did a quick experiment and this can be safely done without any impact on the repo:
> 
> > remove the **array_prepare** from MockFn before upgrade to numpy 2.0

I wanted the CI to confirm this too. But its not possible because pandapower cannot be upgraded to 3.0 in this branch. If anything additional is found it would have to be resolved at the pandapower 3.0 upgradation then.

Lets go with what you suggest.

figueroa1395:
Maybe worth mentioning here, also for reference, that the nightly is failing miserably (https://github.com/PowerGridModel/power-grid-model-io/actions/runs/14256774715/job/39960784060) and it is very likely related to this pandapower issue.

mgovers (approved):

PR #295, opened 2025-04-22
Code quality improvements: Use new enums of dataset and component type

zhen0427:


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -1507,10 +1536,10 @@ def _pp_asym_gens_output(self):
         # TODO: create unit tests for the function
         assert "res_asymmetric_sgen" not in self.pp_output_data
 
-        if "asym_gen" not in self.pgm_output_data or self.pgm_output_data["asym_gen"].size == 0:
+        if "asym_gen" not in self.pgm_output_data or self.pgm_output_data[ComponentType.asym_gen].size == 0:
```
mgovers, line 1539:
```suggestion
        if ComponentType.asym_gen not in self.pgm_output_data or self.pgm_output_data[ComponentType.asym_gen].size == 0:
```

mgovers:
failing sonar cloud seems to be an internal error or another backwards incompatible change
image

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -189,15 +197,15 @@ def test_fill_pgm_extra_info():
 def test_fill_pp_extra_info():
     # Arrange
     converter = PandaPowerConverter()
-    converter.idx_lookup[("line", None)] = pd.Series([102, 103], index=[1, 2])
+    converter.idx_lookup[(ComponentType.line, None)] = pd.Series([102, 103], index=[1, 2])
```
figueroa1395, line 200: Commenting here as it is close, but should we also homogenize in this PR the use of `"trafo"` as `ComponentType.transformer`?

Edit: I see there are other comments that are refered to with colloquial names as well. Do you remember any particular reason for this? I'd say let's just homogenize everything, but seeing that many things would need to be changed, it should probably be done in a separate follow up PR. But may be a good idea to create an issue for it.

Edit 2: I see that the above is probably linked to how things are named on the PP side. If that is the case, would it make sense to make some `Enum`s to keep track of other libraries components as well? Or would that just add too much overhead for us? As those wouldn't be used by the users.

figueroa1395 (commented):
There are some files that have some `str`s left, such as in:
- `test_pandapower_converter_input.py`
- `test_pgm_json_converter.py`
- `base_converter`.py
- `pandapower_converter`.py
- possibly others

In addition, it may also be a good idea to edit the docstrings and replace there the raw `str`s as well.

figueroa1395 (commented):
There are some files that have some `str`s left, such as in:
- `test_pandapower_converter_input.py`
- `test_pgm_json_converter.py`
- `base_converter`.py
- `pandapower_converter`.py
- possibly others

In addition, it may also be a good idea to edit the docstrings and replace there the raw `str`s as well.

tests/unit/converters/test_pandapower_converter_input.py
mgovers: > I see that the above is probably linked to how things are named on the PP side.

indeed

> If that is the case, would it make sense to make some Enums to keep track of other libraries components as well? Or would that just add too much overhead for us? As those wouldn't be used by the users.

it would be good practice, yes. especially to show the difference between pp and pgm specific stuff. i'd do that in a separate PR tho

figueroa1395: Alright, I will make an issue later today to keep track of it.

tests/unit/converters/test_tabular_converter.py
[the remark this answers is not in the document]
zhen0427: This one is resolved. The reason I did not including these changes is that it may cause incompatibility with error message. Now the error message is updated accordingly.

tests/unit/converters/test_pandapower_converter_input.py
figueroa1395: Resolving as the issue was created in https://github.com/PowerGridModel/power-grid-model-io/issues/315

tests/unit/converters/test_tabular_converter.py
```diff
@@ -233,13 +236,13 @@ def test_convert_col_def_to_attribute(
 ):
     with pytest.raises(
         KeyError,
-        match=r"Could not find attribute 'incorrect_attribute' for 'nodes'. \(choose from: id, u_rated\)",
+        match=r"Could not find attribute \'incorrect_attribute\' for \'ComponentType.nodes\'. \(choose from: id, u_rated\)",
```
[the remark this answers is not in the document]
zhen0427: This problem is solved by using component.value in the code.

tests/unit/converters/test_pandapower_converter_input.py
figueroa1395: @petersalemink95 you are right, here some of the components from PP are being named as `ComponentType.something`. We have to be careful with not introducing that convention to not PGM related components. https://github.com/PowerGridModel/power-grid-model-io/issues/315 might come handy then, but I would still say let's do that as a follow up PR, we just have to be extra careful here.

src/power_grid_model_io/converters/pgm_json_converter.py
```diff
@@ -119,7 +120,7 @@ def _parse_dataset(
 
     @staticmethod
     def _parse_component(
-        objects: ComponentList, component: str, data_type: str, extra_info: Optional[ExtraInfo]
+        objects: ComponentList, component: Enum, data_type: Enum, extra_info: Optional[ExtraInfo]
```
mgovers, line 123: ?
```suggestion
        objects: ComponentList, component: ComponentType, data_type: DatasetType, extra_info: Optional[ExtraInfo]
```

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -145,9 +146,9 @@ def _parse_data(self, data: TabularData, data_type: str, extra_info: Optional[Ex
     def _convert_table_to_component(  # pylint: disable = too-many-arguments,too-many-positional-arguments
         self,
         data: TabularData,
-        data_type: str,
+        data_type: Union[str, Enum],
         table: str,
-        component: str,
+        component: Union[str, Enum],
```
mgovers, line 151: what about either `StrEnum` or just directly `DatasetType`/`ComponentType`?

please also use `|` instead of `Union` (python>=3.10), e.g. `str | Enum` if you have to

zhen0427: Because the test may past some fake/wrong datatype and componenttype to this function, like "dummy", so it has to be either str or Enum

src/power_grid_model_io/converters/pgm_json_converter.py
zhen0427: Done

src/power_grid_model_io/converters/tabular_converter.py
mgovers: then fine for now but please use `str | Enum` in the future 😬 it's cleaner and there are actually discussions about deprecating `typing.Union` entirely

mgovers (commented):
LGTM

src/power_grid_model_io/data_stores/json_file_store.py
```diff
@@ -105,5 +105,5 @@ def _validate(self, data: StructuredData) -> None:
             if len(type_names) == 1:
                 type_str = type_names.pop()
             else:
-                type_str = "Union[" + ", ".join(type_names) + "]"
+                type_str = type_str = " | ".join(type_names)
```
nitbharambe, line 108: An error here.

figueroa1395: How so? If we define for instance
```
type_names = ["int", "str", "float"]
```
the old output would be
```
type_str = "Union[" + ", ".join(type_names) + "]"
print(type_str)
# Output: Union[int, str, float]
```
And the new output would be
```
type_str = " | ".join(type_names)
print(type_str)
# Output: int | str | float
```
Which is equivalent, isn't it?

nitbharambe:
```suggestion
                type_str = " | ".join(type_names)
```
This :D

nitbharambe (approved):
- [x] Checked that pandapower's pgm tests pass.

Issue #297, opened 2025-05-08
[BUG] No module named 'power_grid_model._utils'


Bjorn346:
**Describe the bug**

When importing PgmJsonConverter, the error message 'ModuleNotFoundError: No module named 'power_grid_model._utils'' is appearing. The power_grid_model._utils does not exist anymore on that location, since version v1.10.107 of the Power-grid-model, but was moved to power_grid_model._core.utils.

**To Reproduce**

`from power_grid_model_io.converters import PgmJsonConverter`

**Expected behavior**

No error message in Pyhton

**Output from commandline**

Traceback (most recent call last):
  File "c:\*\Calc_MV_voltages.py", line 1, in <module>
    from power_grid_model_io.converters import VisionExcelConverter, PgmJsonConverter
  File "C:\*\.venv\Lib\site-packages\power_grid_model_io\converters\__init__.py", line 9, in <module>     
    from power_grid_model_io.converters.pgm_json_converter import PgmJsonConverter
  File "C:\*\.venv\Lib\site-packages\power_grid_model_io\converters\pgm_json_converter.py", line 16, in <module>
    from power_grid_model._utils import is_nan
ModuleNotFoundError: No module named 'power_grid_model._utils'

**Suggested solution**

Within file "src/power_grid_model_io/converters/pgm_json_converter.py" replace line 16 from 
`from power_grid_model._utils import is_nan` to `from power_grid_model._core.utils import is_nan`

mgovers:
Hi @Bjorn346 , you are completely right. We missed this.

@nitbharambe @Jerry-Jinfeng-Guo for compatibility, this also requires bumping `power-grid`model >= v1.10.x` but it may still be problematic for older pgm-io versions if they version pin on older pgm versions. It may be better to do this on the power-grid-model side:

* re-create `power_grid_model._utils` and import these functions and merge
* then add to pgm-io the requirement `power-grid-model < v1.11`
* then create a new PR on power-grid-model to remove power-grid_model._utils` (to be merged for the `v1.11.0` release)
* create new PR to pgm-io to apply the suggested fix here + bump `power-grid-model >= v1.11`

Alternatively, only do 2 PRs on the PGM-IO side to
* first pin `powe-grid-model < the version before emas PR` and merge and release
* and then apply this fix + bump `power-grid-model >= version with ema's PR`

@Bjorn346 you can maybe also do the fix.

nitbharambe:
[...]

PR #299, opened 2025-05-09
PGM-IO internal import resolution: Copy is_nan functionality from PGM, Step 4

nitbharambe:
**Fixes issue**: https://github.com/PowerGridModel/power-grid-model-io/issues/297

The alternative to Step 4, of copying the is_nan functionality is done. Since this is implemented, we can skip step 2

### DIFF
--- a/src/power_grid_model_io/converters/pgm_json_converter.py
+++ b/src/power_grid_model_io/converters/pgm_json_converter.py
@@ -300,15 +300,15 @@ class PgmJsonConverter(BaseConverter[StructuredData]):
         """
         Determine if the data point is valid
         Args:
-            data: a single scaler or numpy array
+            data: a single scalar or numpy array
 
         Returns:
-            True if all the data points are invalid
+            True when all the data points are invalid
             False otherwise
         """
         nan_func = {
             np.dtype("f8"): lambda x: np.all(np.isnan(x)),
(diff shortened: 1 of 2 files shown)


src/power_grid_model_io/converters/pgm_json_converter.py
```diff
@@ -296,3 +294,21 @@ def _get_first_by(data: List[Dict[str, Any]], field: str, value: Any) -> Optiona
                 return entry
 
         return None
+
+    @staticmethod
+    def _is_nan(data: np.ndarray) -> bool:
+        """
+        Determine if the data point is valid
```
Jerry-Jinfeng-Guo, line 301:
```suggestion
        Determine whether the data point is valid
```

src/power_grid_model_io/converters/pgm_json_converter.py
```diff
@@ -296,3 +294,21 @@ def _get_first_by(data: List[Dict[str, Any]], field: str, value: Any) -> Optiona
...
+        Args:
+            data: a single scaler or numpy array
+
+        Returns:
+            True if all the data points are invalid
```
Jerry-Jinfeng-Guo, line 306:
```suggestion
            True when all the data points are invalid
```

Jerry-Jinfeng-Guo (approved):

PR #300, opened 2025-05-13
Nightly fix: Update nightly for renaming

nitbharambe:
Nightly failed in  https://github.com/PowerGridModel/power-grid-model-io/actions/runs/14986900277/workflow

As per changes in https://github.com/PowerGridModel/power-grid-model-io/pull/296, the nightly file needs to be updated too.
Verify if this renaming is okay or not, since the above PR has additional contents regarding publishing at pypi.

PR #296, opened 2025-04-30
Add trusted publisher

figueroa1395:
Adds trusted publisher as the method to publish the package to PyPI.

- [x] Manual CI run to trigger the action. https://github.com/PowerGridModel/power-grid-model-io/actions/runs/14755601739
- [x] Publish package in Test PyPI. https://test.pypi.org/project/power-grid-model-io/#history

PR #301, opened 2025-05-15
Nightly build fail: Move publishing out of nightly

nitbharambe:
This was attempted in #296 and relates to changes in #300

In #300, it was missed that we need to move publishing part out of nightly, but its present in ci.yml.

The diff is huge, the change is simple. Intention is to create a common job of `build-test-release` which is ran in `ci` and in `nightly`. (The way it was before.)
Now ci.yml is left with `build-test-release` jobs and `publish`.

- [x] Manual run of nightly: https://github.com/PowerGridModel/power-grid-model-io/actions/runs/15185232639 (Missed this last time)


.github/workflows/build-test-release.yml
```diff
@@ -0,0 +1,208 @@
...
+          name: power-grid-model-io
+          path: wheelhouse/
+
+      - name: Install built wheel file
+        run: pip install power-grid-model-io[dev]==${{ needs.build-python.outputs.version }} --find-links=wheelhouse
+
+      - name: Validation tests
+        run: pytest tests/validation --no-cov --verbose
+
+  github-release:
```
nitbharambe, line 171: Shall we move `github-release` to ci.yml as well? 
I don't understand the reasoning behind running this with `inputs.create_release`=false. 

I ask because same is done in PGM repo.

mgovers: > Shall we move github-release to ci.yml as well?

we want to homogenize as much as possible between PGM-IO and PGM.

> I don't understand the reasoning behind running this with inputs.create_release=false.

i think that that is a topic for a whole other discussion. Let's first focus on making it work again

mgovers (approved):

figueroa1395 (approved):

Issue #289, opened 2025-01-22
[FEATURE] Enable testing for python 3.13


figueroa1395:
As found in [0310c66](https://github.com/PowerGridModel/power-grid-model-io/pull/288/commits/0310c66ab961904afdbb422f7b6abcf90bb23ee5), even though `power-grid-model-io` supports Python 3.13, one of its dependencies, namely `pandapower`, does not. However, since this is a [dev] dependency and it's only used for testing purposes, we can solve it separately, i.e. with this issue. Furthermore, this issue is about enabling python 3.13 for testing in the CI.

### Possible solutions

1. One possible way to go, and assuming `pandapower` doesn't support Python 3.13 soon, is to split the testing actions in two parts each: One for everything related with this dependency and the other for the rest. The first action, would be for Python 3.11 and 3.12 only, whereas the second would be tested in all supported Python versions available.
2. Another way, is to possibly wait until this is fixed on the dependency side, and only then upgrade the testing on the CI to Python 3.13.

My proposed solutions are not exhaustive, so feel free to add anything you might consider appropriate.

PR #302, opened 2025-05-15
Fix sonar coverage

nitbharambe:
image

Sonar scan shows 0.0 coverage and this warning is shown.

Relates to https://github.com/PowerGridModel/power-grid-model/pull/1002

PR #303, opened 2025-05-19
dependencies: compatibility pandapower 2 and 3 and numpy 1 and 2

mgovers:
This removes the `pandapower<3` requirement and therefore we can now also test using `numpy>=2` (instead of only using `numpy<2`. Note that we can also apply such a change to the conda-forge feedstock repo

dependencies checked:

| Python | pandapower | numpy  | pandas | scipy  | Compatibility checked? |
| :----: | :--------: | :----: | :----: | :----: | :--------------------: |
| 3.12   | 3.0.0      | 2.2.6  | 2.2.3  | 1.13.1 | ✅                     |
| 3.12   | 3.3.0      | 1.26.4 | 2.2.3  | 1.13.1 | ✅                     |
| 3.12   | 2.14.11    | 1.26.4 | 2.2.3  | 1.13.1 | ✅                     |

We cannot test Python 3.13 yet due to #289 , but it does provide a possible solution for it.


tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -1462,14 +1469,19 @@ def test_create_pgm_input_three_winding_transformers__tap_dependent_impedance()
...
 
-    # Act/Assert
-    with pytest.raises(RuntimeError, match="not supported"):
-        converter._create_pgm_input_three_winding_transformers()
+        converter = PandaPowerConverter()
+        converter.pp_input_data = {k: v for k, v in pp_net.items() if isinstance(v, pd.DataFrame)}
+
+        # Act/Assert
+        with pytest.raises(RuntimeError, match="not supported"):
+            converter._create_pgm_input_three_winding_transformers()
```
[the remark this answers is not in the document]
mgovers: yes because there now is an additional if/else statement concerning pandapower version

mgovers:
we need to decide on what to do with the validation cases. https://github.com/e2nIEE/pandapower/issues/2557 is still alive.

One possible way forward is to run the validation tests on `pandapower<3`, so that at least the software package is compatible - just untested. @nitbharambe thoughts?

nitbharambe:
Let us investigate if we can serialize the network results and validate against it instead of running a PF in pandapower. PGMIO eould then function only as a converter.

mgovers:
> return net

done by explicitly adding `pandapower<3` data to the validation tests

mgovers:
Failing sonar-cloud build is out of scope of this PR and is done in #302

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -1462,14 +1469,19 @@ def test_create_pgm_input_three_winding_transformers__tap_dependent_impedance()
...
```
figueroa1395, line 1485: Nitpick:
```suggestion
        with pytest.raises(RuntimeError, match=r"not supported"):
```

Probably at other places as well.

figueroa1395 (commented):
LGTM, only minor nitpick.

Side note: The large change in files/lines, comes from the fact that we don't run the pandapower calculation directly for the validation cases (due to the pandapower 3 missmatch), but we rather ran it with pandapower 2, and saved the resulting files to be used as the data here for direct validation.

figueroa1395 (commented):
LGTM, only minor nitpick.

Side note: The large change in files/lines, comes from the fact that we don't run the pandapower calculation directly for the validation cases (due to the pandapower 3 missmatch), but we rather ran it with pandapower 2, and saved the resulting files to be used as the data here for direct validation.

tests/unit/converters/test_pandapower_converter_input.py
figueroa1395: Github messed up... I only meant to comment about the raw string part...

figueroa1395 (approved):

PR #305, opened 2025-05-22
ci: enable python 3.13 tests by skipping pandapower stuffs

### LINKED ISSUE #289: [FEATURE] Enable testing for python 3.13

mgovers:
Fixes #289 

pandapower is not compatible with a recent version of scipy and as a result not with python 3.13. We still want to test python 3.13, though. We decided to disable coverage for the pandapower converter for the time being (follow-up issue to be created) to reenable this. We also had to add a subset dependency group in the pyproject.toml that excludes pandapower while maintaining backwards compatibility.

@nitbharambe can you please review?

c.c. @TonyXiang8787

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,18 +39,20 @@ dependencies = [
 dynamic = ["version"]
 
 [project.optional-dependencies]
-dev = [
-    "pandapower>2.11.1",
+dev-bare = [  # pandapower 3.0.0 is not compatible with Python 3.13. pandapower 3.1 will be.
     "pre-commit",
     "pylint",
     "pytest",
     "pytest-cov",
     "pydantic>2", # Used in unit tests
-    "numba",
+]
+dev = [
+    "power-grid-model-io[dev-bare]",
+    "pandapower>2.11.1, <3.1", # pandapower 3.0.0 is not compatible with Python 3.13. pandapower 3.1 will be.
 ]
 examples = [
     "power-grid-model>1.9.80",
-    "pandapower>2.11.1, <3.0",
+    "pandapower>2.11.1, <3.0", # pandapower 3.0.0 has differences in results cfr. https://github.com/e2nIEE/pandapower/issues/2557
     "pyarrow", # Pyarrow support for Python 3.12 scheduled for 14.0.0: https://github.com/apache/arrow/issues/37880
 ]
 doc = [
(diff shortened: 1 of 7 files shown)


nitbharambe:
Good idea with this PR.
Lets also have this decision documented? A separate title in release strategy would do right?

pyproject.toml
```diff
@@ -39,18 +39,20 @@ dependencies = [
...
 [project.optional-dependencies]
-dev = [
-    "pandapower>2.11.1",
+dev-bare = [  # pandapower 3.0.0 is not compatible with Python 3.13. pandapower 3.1 will be.
     "pre-commit",
     "pylint",
     "pytest",
     "pytest-cov",
     "pydantic>2", # Used in unit tests
-    "numba",
```
nitbharambe, line 49: This should not have been included, good catch

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -35,7 +35,7 @@
 
 
 # pylint: disable=too-many-instance-attributes
-class PandaPowerConverter(BaseConverter[PandaPowerData]):
+class PandaPowerConverter(BaseConverter[PandaPowerData]):  # pragma: no cover
```
nitbharambe, line 38: Does this ignore the entire class? Then the coverage gets ignored for all versions too right?

mgovers: let's just hope that we do not need to wait so long before pp fixes their stuff

PR #312, opened 2025-06-05
CI: add no label check

Jerry-Jinfeng-Guo:
Mirror of https://github.com/PowerGridModel/power-grid-model/pull/1010


figueroa1395:
@mgovers Can you please remove the required `check-blocking-labels` check?

figueroa1395:
> We rename the branch setting post merge right?
> 
> * [ ]  Renamed the job name to check-pr-labels in branch settings in GH.

We need to remove the check now the old check in the settings and add the new one, manually. It has to be done now or this won't merge as the checks will never go through. However, I don't have the permissions to do so.

PR #313, opened 2025-06-10
Fix ci: github actions requires ci-passed + bump pre-commit + fix mypy

mgovers:
cfr. https://github.com/PowerGridModel/power-grid-model/pull/1015


nitbharambe:
Branch name does not match changes @mgovers

PR #320, opened 2025-08-24
Pandapower converter: fix 3-Ph transformer loading Calculation

### LINKED ISSUE #319: [BUG] *Pandapower 3-Ph transformer loading Calculation*
**Value Error and Invalid loading percent**

When running asymmetric loadflow in pandapower using runpp_pgm and with option trafo_loading="Current", a value error occurs in pandapower_converter.py at _pp_trafos_output_3ph function because "i_from" and "i_to" have shape (n,3) while "u1" and "u2" have shape (n,). Moreover, the loading calculation is invalid; for phase wise loading, the voltage "u" which is line-to-line needs to be converted to line-to-phase by dividing with sqrt(3), moreover, the phase power needs to be divided with sn_phase = sn / 3.

### LINKED ISSUE #321: [IMPROVEMENT] Pandapower_convert gives NaN for tan1, the positive-sequence shunt loss factor
**Describe the problem**

When `c_nf_per_km` line shunt capacitance equals to 0 in PandaPower, PGM-io sets `tan1` to NaN due to division by zero at line 426 in pandapower_converter.py. However, assert_valid_input_data in PGM does not accept NaN values for `tan1`. 

**Improvement proposal**

When `c_nf_per_km` line shunt capacitance equals to 0 in PandaPower, setting `tan1` to 0. This matches the the fact that `c_nf_per_km = 0` means no shunt capacitance, and therefore no `tan1`, the positive-sequence shunt loss factor.

Details see the attached Jupyter notebook.

[requirements.txt](https://github.com/user-attachments/files/22008095/requirements.txt)
[ieee14_conversion.ipynb](https://github.com/user-attachments/files/22008094/ieee14_conversion.ipynb)

furqan463:
Fixes #319
Fixes #321 

Extended "u1" and "u2" arrays from (n,) dimensions to (n,3) dimensions to remove value error.
multiplied phase loading percent with sqrt(3), (i * v / sqrt(3)) / (sn / 3) --> sqrt(3) * (i * v) / sn
divided total loading percent with sqrt(3) as v_ph = v / sqrt(3). 

### Electrical Mathematics:
Power per Phase = V_n * I_line_to_line                               # Basic Definition
Capacity per Phase = Sn / 3 
Loading per Phase = Power per Phase / Capacity per Phase

ui = i_line_to_line * V_line_to_line
Loading per Phase  = [ui / sqrt(3)]  / [Sn  /  3]
                                = [3 / sqrt(3)] * [ui]  /  [Sn]
                                = sqrt(3) * ui / Sn

Total Power = sum[Power_Phase_a, Power_Phase_b, Power_Phase_c]
                    = sum[ ui_a / sqrt(3), ui_b / sqrt(3), ui_c / sqrt(3)]
                    = [ sum(ui_a, ui_b, ui_c) / sqrt(3) ]
Total Loading = [sum(ui_a, ui_b, ui_c) / sqrt(3)] / Sn 

Looking at it from another perspective which is more relevant in case of trafo_loading="current"
loading_per_phase = phase_current(line_to_line) / rated_current
rated_current (I_rated) = Sn / (sqrt(3) x V_line_to_line)         #As for 3-Ph power system S = sqrt(3) * V * I
loading_per_phase = I_ph / I_rated
                               = I_ph / [Sn / (sqrt(3) x V_line_to_line)]
                               = sqrt(3) * [I_ph * V_line_to_line] / Sn

### Checks before merge

* [x] Review the mathematics
* [x] Reproduce; verify that this fixes the issue
* [x] Code review
* [x] Add tests
* [x] Pass CI (Sonar Cloud may fail with HTML error code 403. That is not a problem with this PR)

### DIFF
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -360,9 +360,9 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         Furthermore, creates a global node lookup table, which stores nodes' voltage magnitude per unit and the voltage
         angle in degrees
         """
-        #TODO create output_data_3ph for remaining components
-        #Although Pandapower itself did not implmenet res_shunt_3ph
-        #Since results are avaiable in PGM output, these should be converted.
+        # TODO create output_data_3ph for remaining components
+        # Although Pandapower itself did not implmenet res_shunt_3ph
+        # Since results are avaiable in PGM output, these should be converted.
         self._pp_buses_output_3ph()
         self._pp_lines_output_3ph()
         self._pp_ext_grids_output_3ph()
@@ -379,7 +379,7 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         Returns:
             a power-grid-model structured array for the Node component
         """
-        #TODO handle out-of-service buses, either here or in get_switch_states
+        # TODO handle out-of-service buses, either here or in get_switch_states
         pp_busses = self.pp_input_data["bus"]
 
         if pp_busses.empty:
--- a/tests/data/pandapower/pp_v2_net_3ph_output_current_loading.json.license
+++ b/tests/data/pandapower/pp_v2_net_3ph_output_current_loading.json.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+
+SPDX-License-Identifier: MPL-2.0


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2017,12 +2017,21 @@ def _pp_trafos_output_3ph(self):  # pylint: disable=too-many-statements
...
-            ui_to = pgm_output_transformers["i_to"] * pgm_input_transformers["u2"]
-            loading_a_percent = np.maximum(ui_from[:, 0], ui_to[:, 0]) / pgm_input_transformers["sn"]
-            loading_b_percent = np.maximum(ui_from[:, 1], ui_to[:, 1]) / pgm_input_transformers["sn"]
-            loading_c_percent = np.maximum(ui_from[:, 2], ui_to[:, 2]) / pgm_input_transformers["sn"]
-            loading = np.maximum(np.sum(ui_from, axis=1), np.sum(ui_to, axis=1)) / pgm_input_transformers["sn"]
+            # ui_from = pgm_output_transformers["i_from"] * pgm_input_transformers["u1"]
+            # ui_to = pgm_output_transformers["i_to"] * pgm_input_transformers["u2"]
+            # since "i_from" and "i_to" are (n, 3) arrays while "u1" and "u2" are (n,) arrays, valueError is generated during broadcost
+            ui_from = pgm_output_transformers["i_from"] * np.tile(pgm_input_transformers["u1"][:, None], (1,3))
+            ui_to = pgm_output_transformers["i_to"] * np.tile(pgm_input_transformers["u2"][:, None], (1,3))
```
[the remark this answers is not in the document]
furqan463: Yes you're right sir. Removed tiling.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2017,12 +2017,19 @@ def _pp_trafos_output_3ph(self):  # pylint: disable=too-many-statements
...
+            ui_to = pgm_output_transformers["i_to"] * pgm_input_transformers["u2"][:, None]
+            # for phase wise loading, sn_ph = sn / 3, v_n = v / sqrt(3), so (i * u / sqrt(3)) / (sn / 3)
+            # ==> sqrt(3) * i * u / sn
+            loading_a_percent = np.sqrt(3) * np.maximum(ui_from[:, 0], ui_to[:, 0]) / pgm_input_transformers["sn"]
+            loading_b_percent = np.sqrt(3) * np.maximum(ui_from[:, 1], ui_to[:, 1]) / pgm_input_transformers["sn"]
+            loading_c_percent = np.sqrt(3) * np.maximum(ui_from[:, 2], ui_to[:, 2]) / pgm_input_transformers["sn"]
+            # for total loading, phase wise powers are be calculated using v_n instead of v.
+            loading = (
+                np.maximum(np.sum(ui_from, axis=1), np.sum(ui_to, axis=1)) / np.sqrt(3)
+            ) / pgm_input_transformers["sn"]
```
nitbharambe, line 2032:
```suggestion
            loading = (
                np.maximum([loading_a_percent, loading_b_percent, loading_c_percent])
```
This would be equivalent to what pandapower gives in https://github.com/e2nIEE/pandapower/blob/7b7ab33a8a9cc1b9802e168dfeea824eeb2ef266/pandapower/results_branch.py#L461-L465 right?

furqan463:
@mgovers I've updated the description above elaborating mathematics involved. Moreover, thanks to @nitbharambe proposed changes match with existing code in PandaPower. I'm attaching a small code to reproduce the errors as well as testing the proposed changes. I'm trying to create unit tests, but while working on it, I've found some bugs with existing validation data.
[pgm_test.py](https://github.com/user-attachments/files/21973364/pgm_test.py)

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -126,6 +127,74 @@ def test_generate_output_3ph():  # TODO: REMOVE THIS FUNCTION
         json_converter.save(data=output_data_asym, extra_info=extra_info)
 
 
+def test_output_trafos_3ph__power__with_comparison():
+    import numpy as np
+
+    def check_result(net):
+        v_hv = net.trafo.vn_hv_kv
+        v_lv = net.trafo.vn_lv_kv
```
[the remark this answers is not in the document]
mgovers: won't fix

furqan463:
@mgovers I see you've added the minimal test and it's working with rtol=0.04 but there's a problem with this test and the reason why I did not make a validation test to compare pp and pgm results. In fact the reason I turned to pgm for 3ph and happen to find this bug is because PandaPower do not properly handle Dyn transformer in unbalanced condition. All it does is remove zero sequence from load, while actually hv current of each phase is vector difference of lv currents of two related phases. In this case the imbalance in the load is minimal so it's working with rtol=0.04, if we use balanced load, the test passes with rtol=0.004 as well which I'm going to do in next commit. But conceptually it's not right. We cannot validate pgm results against pp results in unbalanced 3ph analysis with delta-start transformation. And lastly in my opinion comparison of results of pp and pgm is not in the scope of converter.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -422,11 +422,12 @@ def _create_pgm_input_lines(self):
...
-        pgm_lines["tan1"] = (
-            self._get_pp_attr("line", "g_us_per_km", expected_type="f8", default=0)
-            / c_nf_per_km
-            / (2 * np.pi * self.system_frequency * 1e-3)
+        pgm_lines["tan1"] = np.divide(
+            self._get_pp_attr("line", "g_us_per_km", expected_type="f8", default=0),
+            c_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
+            where=np.not_equal(c_nf_per_km, 0.0),
         )
+        pgm_lines["tan1"][np.equal(c_nf_per_km, 0.0)] = 0.0
```
mgovers, line 430: can you please add a test case for this?

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -435,11 +436,12 @@ def _create_pgm_input_lines(self):
...
-        pgm_lines["tan0"] = (
-            self._get_pp_attr("line", "g0_us_per_km", expected_type="f8", default=0)
-            / c0_nf_per_km
-            / (2 * np.pi * self.system_frequency * 1e-3)
+        pgm_lines["tan0"] = np.divide(
+            self._get_pp_attr("line", "g0_us_per_km", expected_type="f8", default=0),
+            c0_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
+            where=np.not_equal(c0_nf_per_km, 0.0),
         )
+        pgm_lines["tan0"][np.equal(c0_nf_per_km, 0.0)] = 0.0
```
mgovers, line 444: can you please add a test case for this?

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -126,6 +127,88 @@ def test_generate_output_3ph():  # TODO: REMOVE THIS FUNCTION
...
+            actual.res_trafo_3ph.loading_c_percent, expected.res_trafo_3ph.loading_c_percent, rtol=rtol
+        )
+
+    pgm_net = pp_net_3ph_minimal_trafo()
+    pp_net = pp_net_3ph_minimal_trafo()
+    pp.runpp_pgm(pgm_net, symmetric=False)
+    pp.runpp_3ph(pp_net)
+    check_result(pgm_net)
+    check_result(pp_net)
+    compare_result(pgm_net, pp_net, rtol=0.004)
```
mgovers, line 209: can you maybe add 2 test cases?

1. balanced => `rtol=0.004`
2. unbalanced => `rtol=0.04` and also add a check that the test fails when `rtol=0.004`, as well as a comment that explains why (namely that pandapower does not model the unbalanced case correctly)

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -164,6 +247,9 @@ def test_attributes(output_data: Tuple[PandaPowerData, PandaPowerData], componen
     pd.testing.assert_series_equal(actual_values, expected_values, atol=5e-4, rtol=1e-4)
 
 
+# The following test only works for those components where valid data is returned by
+# load_and_convert_pgm_data_3ph. since this is failing for trafo_output_3ph (returning
+# from first "if", this function's output is not being tested currently.
```
mgovers, line 252: a comment to state the fact that there is an action item needed

```suggestion
# TODO(furqan463)
# The following test only works for those components where valid data is returned by
# load_and_convert_pgm_data_3ph. since this is failing for trafo_output_3ph (returning
# from first "if", this function's output is not being tested currently.
```

mgovers (commented):
hi @furqan463,

Since you are actively picking up this PR again, the team has decided to treat this as a regular external contribution PR again and only merge it once all testing is fully in place.

Please let me know if you need help - if not, I'll be on stand-by.

nitbharambe:
> @mgovers I see you've added the minimal test and it's working with rtol=0.04 but there's a problem with this test and the reason why I did not make a validation test to compare pp and pgm results. In fact the reason I turned to pgm for 3ph and happen to find this bug is because PandaPower do not properly handle Dyn transformer in unbalanced condition. All it does is remove zero sequence from load, while actually hv current of each phase is vector difference of lv currents of two related phases. In this case the imbalance in the load is minimal so it's working with rtol=0.04, if we use balanced load, the test passes with rtol=0.004 as well which I'm going to do in next commit. But conceptually it's not right. We cannot validate pgm results against pp results in unbalanced 3ph analysis with delta-start transformation. And lastly in my opinion comparison of results of pp and pgm is not in the scope of converter.

> On one hand, we do want to compare the two (as long as the difference is explained, it serves as a sanity check/validation: PP and PGM should be similar but not identical).

> so it is possible that the tests will break in the future as well, but at least, we'll be notified and can act accordingly when it happens.

Agreed with these! The validation test here and rtols should not be used for modelling comparison. Hence, we should be less stringent on the rtol setting here if test fails because of it (Now or in future.)

We had earlier also thought of having a default vs custom rtol for specific attribute, but that was ruled out. We can do it if more such cases occur (Not because it would enable finer validation, but to clean up code.).

tests/validation/converters/test_pandapower_converter_output.py
furqan463: > a comment to state the fact that there is an action item needed

This is already resolved, as data for some components was missing because there was no pp_data available. With Addition of this line "converter.load_input_data(load_validation_data_3ph(), make_extra_info=False)" in load_and_convert_pgm_data_3ph, the issue is resolved. so the comments can be removed.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2017,38 +2022,35 @@ def _pp_trafos_output_3ph(self):  # pylint: disable=too-many-statements
...
+            loading_c_percent = np.maximum(
+                pgm_output_transformers["s_from"][:, 2],
+                pgm_output_transformers["s_to"][:, 2],
+            ) / (pgm_input_transformers["sn"] / 3)
         else:
             raise ValueError(f"Invalid transformer loading type: {str(self.trafo_loading)}")
 
+        # PGM returns the average loading over the trafos, but PandaPower returns the maximum of the per-phase loading.
+        # To make it consistent with PandaPower, overall loading will be calculated as max of above 3.
+        loading = np.maximum(np.maximum(loading_a_percent, loading_b_percent), loading_c_percent)
```
mgovers, line 2052: should the line loading also be updated? or is that unaffected?

tests/validation/converters/test_pandapower_converter_input.py
```diff
@@ -116,3 +116,8 @@ def test_extra_info__serializable(extra_info):
 
     # Assert
     json.dumps(actual, cls=JsonEncoder)  # expect no exception
+
+def test_pgm_input_lines__cnf_zero():
+    pp_net = pp_net_3ph_minimal_trafo()
+    pp_converter = PandaPowerConverter()
```
mgovers, line 122: this test also passes with the original. maybe you can try to catch the division by zero warning? there's functionality for that in pytest

or you can patch the `np.divide` and assert that it has been called exactly once and that the divisor does not have `0` components.

to see whether your test is good (does what it should), you can try to temporarily checkout the `src/` directory to the original one (do not commit!!!) and run the test.

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -126,6 +131,100 @@ def test_generate_output_3ph():  # TODO: REMOVE THIS FUNCTION
...
+    # Symmetric Load
+    pgm_net.asymmetric_load.loc[:, ['p_a_mw', 'p_b_mw', 'p_c_mw']] = 0.2
+    pgm_net.asymmetric_load.loc[:, ['q_a_mvar', 'q_b_mvar', 'q_c_mar']] = 0.05
+    pp_net.asymmetric_load.loc[:, ['p_a_mw', 'p_b_mw', 'p_c_mw']] = 0.2
+    pp_net.asymmetric_load.loc[:, ['q_a_mvar', 'q_b_mvar', 'q_c_mar']] = 0.05
+    pp.runpp_pgm(pgm_net, symmetric=False)
+    pp.runpp_3ph(pp_net)
+    check_result(pgm_net)
+    check_result(pp_net)
+    compare_result(pgm_net, pp_net, rtol=0.005)
```
mgovers, line 225: 👍

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -126,6 +131,100 @@ def test_generate_output_3ph():  # TODO: REMOVE THIS FUNCTION
...
+        )
+
+    pgm_net = pp_net_3ph_minimal_trafo()
+    pp_net = pp_net_3ph_minimal_trafo()
+    # Asymmetric Load
+    pp.runpp_pgm(pgm_net, symmetric=False)
+    pp.runpp_3ph(pp_net)
+    check_result(pgm_net)
+    check_result(pp_net)
+    compare_result(pgm_net, pp_net, rtol=0.04)
```
mgovers, line 214:
```suggestion
    # there is a known difference between the modeling in PGM and PP
    # so the rtol is higher, but still reasonably small.
    compare_result(pgm_net, pp_net, rtol=0.04)
```

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -164,7 +263,7 @@ def test_attributes(output_data: Tuple[PandaPowerData, PandaPowerData], componen
     pd.testing.assert_series_equal(actual_values, expected_values, atol=5e-4, rtol=1e-4)
 
 
-@pytest.mark.parametrize(("component", "attribute"), component_attributes_df(load_and_convert_pgm_data_3ph()))
+@pytest.mark.parametrize(("component", "attribute"), component_attributes_df(load_and_convert_pgm_data_3ph()))
```
mgovers, line 266: i will run the formatter for you once you've notified us that the current state is good.

tests/validation/converters/test_pandapower_converter_output.py
mgovers: resolved

src/power_grid_model_io/converters/pandapower_converter.py
mgovers: resolved

mgovers (commented):
@furqan463 looks quite good. A couple small remarks left, but I think the main piece of the code is pretty much finished.

tests/validation/converters/test_pandapower_converter_input.py
furqan463: I started creating this test but then found out we're not testing both cases of "trafo_loading" option so this test left unfinished. Will complete and commit.

furqan463:
@mgovers I'm done now. There's one unittest failing that I'm not sure how to resolve. As I understand the test check exactly how we calculate "tan1" in pgm_input_line, but I changed the formula to avoid Nan values.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -1704,14 +1708,16 @@ def join_currents(table: str, bus_name: str, i_name: str) -> pd.DataFrame:
...
 
         # For et=b, ie bus to bus switches, links are created. get result from them
         if not links_absent:
             links = self.pgm_output_data[ComponentType.link]
             # For links, i_from = i_to = i_ka / 1e3
             link_ids = self._get_pp_ids("switch", links["id"], "b2b_switches")
             pp_switches_output.loc[link_ids, "i_ka"] = links["i_from"] * 1e-3
+        in_ka = self.pp_input_data["switch"]["in_ka"].values
+        pp_switches_output["loading_percent"] = np.nan
+        pp_switches_output["loading_percent"] = np.divide(pp_switches_output["i_ka"], in_ka, where=in_ka != 0)
```
mgovers, line 1720: I will double-check tomorrow that this does what it should, or whether it should instead be something along the lines of

```suggestion
        pp_switches_output["loading_percent"] = np.nan
        pp_switches_output["loading_percent"][in_ka != 0] = np.divide(pp_switches_output["i_ka"], in_ka, where=in_ka != 0)
```

tests/validation/converters/test_pandapower_converter_input.py
```diff
@@ -116,3 +116,16 @@ def test_extra_info__serializable(extra_info):
...
+def test_pgm_input_lines__cnf_zero():
+    pp_network = pp_net_3ph_minimal_trafo()
+    pp_converter = PandaPowerConverter()
+    pp_network.line.c_nf_per_km = 0
+    data, _ = pp_converter.load_input_data(pp_network)
+    assert data[ComponentType.line]["tan1"] == 0
+    pp_network.line.c_nf_per_km = 0.001
+    pp_network.line.c0_nf_per_km = 0
+    data, _ = pp_converter.load_input_data(pp_network)
+    assert data[ComponentType.line]["tan0"] == 0
```
mgovers, line 131: shouldn't it do the opposite? as in: check that it does **not** raise this warning anymore?

I will have a quick look tomorrow morning

mgovers: done

mgovers:
@nitbharambe looks like this PR is now in a good-to-go state. Since I've also contributed significantly, can you please review?

mgovers:
OH apparently, CI is still failing; investigating.

mgovers (commented):
found the reason for the remaining validation tests.

To run the validation tests locally, please either use a test extension or run `pytest tests/validation` rather than just `pytest`, as the validation tests are excluded from `pytest` by default.

tests/validation/converters/test_pandapower_converter_output.py
```diff
@@ -26,6 +26,7 @@
 PGM_ASYM_OUTPUT_FILE = PGM_PP_TEST_DATA / "pgm_asym_output_data.json"
 PP_V2_NET_OUTPUT_FILE = PGM_PP_TEST_DATA / "pp_v2_net_output.json"
 PP_V2_NET_3PH_OUTPUT_FILE = PGM_PP_TEST_DATA / "pp_v2_net_3ph_output.json"
+PP_V2_NET_3PH_OUTPUT_FILE_CURRENT_LOADING = PGM_PP_TEST_DATA / "pp_v2_net_3ph_output_current_loading.json"
```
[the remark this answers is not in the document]
furqan463: done

furqan463:
@mgovers please check now, I've updated the data file.
TODOs added for future.

mgovers:
> @mgovers please check now, I've updated the data file. TODOs added for future.

Great! @nitbharambe and I will have a look on Monday.

* [ ] concerninng the license file for the json data, you can copy-paste one of the other files such that `foo.json` has a file `foo.json.license`. The `.license` files should be the same across the codebase.
* [ ] please install the dev dependencies and run `pre-commit run --all-files` to run the formatter, linter and type checker. It will also run a check that checks whether you added the license file correctly.

tests/validation/converters/test_pandapower_converter_input.py
```diff
@@ -66,29 +72,35 @@ def test_input_data(input_data: Tuple[SingleDataset, SingleDataset]):
     """
     Unit test to preload the expected and actual data
     """
-    # Arrange
-    actual, expected = input_data
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
nitbharambe, line 76: Its much cleaner to use `@pytest.mark.filterwarnings("error")` decorator instead

mgovers: 100% i'll do this change

furqan463: But It'll generate errors for other warnings as well. Shouldn't we filter only "RuntimeWarning" as error.

mgovers: @furqan463 you're right and they are completely unrelated to this PR. I had an offline chat with @nitbharambe yesterday and we decided to scope this improvement out of this PR. We will make a good-first-issue follow-up ticket of this.

nitbharambe (approved):
LGTM, Thanks for the contribution @furqan463!

mgovers:
Anytime! Thank you very much for your contribution. You've been a great contributor. If you like, could you please share your experiences and maybe also any suggestions for improvements?

Issue #322, opened 2025-09-01
[BUG] using the exclude_empty filter doesnt filter out rows


vincentkoppen:
**Describe the bug**
The VisionConverter with a yaml that uses the `exclude_empty` filter, doesn't filter out any rows. 
Specifically if the column is a float column.

Currently the `exclude_empty` filter corrects for a `pd.Series` value of `row[col]` áfter it has called `has_value`. 
However `has_value` doesnt work correctly for pd.Series, since it expects a standard python object. 

**Input Data Validity**

Did you check the validity of input data using the helper function? 
yes
[Examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Validation%20Examples.html)

**Data Dump**
-
Please dump the input (and maybe also the output) data into json files using the helper function, if applicable
[Examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Make%20Test%20Dataset.html#export-to-json)

**To Reproduce**

- Run the Vision Converter with an yaml that uses filters exclude empty
- The filter doesn't work

**Expected behavior**

- The filter filters out empty rows of a column

**Screenshots**

-

prissporton10:
I think this will fix it.
https://droplink.digital/file/3f95ce00b2c01e917344d0379d23bfcf

mgovers:
Fixed by https://github.com/PowerGridModel/power-grid-model-io/pull/323

PR #323, opened 2025-09-01
fix(exclude_empty): correcty handly pd.Series row[col] values

vincentkoppen:
**Fixes issue**: #322 

- Within the `exclude_empty` filter, we convert van `pd.Series` to a value before we call the `has_value` function instead of after. The `has_value` uses `isinstance` for certain checks, so a float check will otherwise fail. 
- I also updated `exclude_value` and added some tests, this function did work but I do think it's nice to transform as early as possible. 

- 

- 
- 
- ..

### Checks

- [ ] If changes are related to CI/CD, are they verified via a manual run on this branch?


mgovers (approved):
LGTM

PR #326, opened 2025-09-10
Pandapower_converter bug fix: Resolve bug with broadcasting

furqan463:
### Changes proposed in this PR include:

The changes proposed in PR #320 fixed known issues but created a bug in pgm_input_line function of pandapower_converter. in case of no data in column c0_nf_per_km, _get_attr_pp returns a single element array as it states broadcasting is responsibility of caller, but broadcasting was not handled, which creates error in assignment statemetns.

I've added another test to the /validation/converter/test_pandapower_converter_input.py, copied from pandapower tutorial power-grid-model_powerflow.ipynb, which fails on existing code.

PR #327, opened 2025-09-12
Fix: Remove Redundent Initialization to resolve reliability issue on Pypi.

furqan463:

Removed redundant initialisation lines that were causing reliability issue on Pypi concerning array values being set twice.

### DIFF
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -426,7 +426,6 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         pgm_lines["x1"] = self._get_pp_attr("line", "x_ohm_per_km", expected_type="f8") * multiplier
         pgm_lines["c1"] = c_nf_per_km * length_km * parallel * 1e-9
         # The formula for tan1 = R_1 / Xc_1 = (g * 1e-6) / (2 * pi * f * c * 1e-9) = g / (2 * pi * f * c * 1e-3)
-        pgm_lines["tan1"] = 0.0
         pgm_lines["tan1"] = np.divide(
             self._get_pp_attr("line", "g_us_per_km", expected_type="f8", default=0),
             c_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
@@ -440,7 +439,6 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         pgm_lines["r0"] = self._get_pp_attr("line", "r0_ohm_per_km", expected_type="f8", default=np.nan) * multiplier
         pgm_lines["x0"] = self._get_pp_attr("line", "x0_ohm_per_km", expected_type="f8", default=np.nan) * multiplier
         pgm_lines["c0"] = c0_nf_per_km * length_km * parallel * 1e-9
-        pgm_lines["tan0"] = 0.0
         pgm_lines["tan0"] = np.divide(
             self._get_pp_attr("line", "g0_us_per_km", expected_type="f8", default=0),
             c0_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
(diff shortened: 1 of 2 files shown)


mgovers:
> Can't we run all those tests on this PR before releasing on Pypi.

Unfortunately, SonarQube Cloud did not catch this on your previous PR because it skipped the analysis step because you created the PR from a fork; see also https://github.com/PowerGridModel/power-grid-model-io/blob/main/.github/workflows/sonar.yml#L51

image

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -426,7 +426,6 @@ def _create_pgm_input_lines(self):
         pgm_lines["x1"] = self._get_pp_attr("line", "x_ohm_per_km", expected_type="f8") * multiplier
         pgm_lines["c1"] = c_nf_per_km * length_km * parallel * 1e-9
         # The formula for tan1 = R_1 / Xc_1 = (g * 1e-6) / (2 * pi * f * c * 1e-9) = g / (2 * pi * f * c * 1e-3)
-        pgm_lines["tan1"] = 0.0
```
mgovers, line 429: i think your initial suggestion was actually correct, but I do agree with SonarQube that it would cause unnecessary assignments. The difficulty here lies in the fact that only valid values are broadcasted, which SonarQube is not able to understand, apparently.

Instead, maybe you can do something along the lines of

```python
invalid = np.isclose(c_nf_per_km, 0.0)
pgm_lines["tan1"] = np.divide(
    self._get_pp_attr("line", "g_us_per_km", expected_type="f8", default=0),
    c_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
    where=np.logical_not(invalid),
)
pgm_lines["tan1"][invalid] = 0.0
```

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -440,7 +439,6 @@ def _create_pgm_input_lines(self):
         pgm_lines["r0"] = self._get_pp_attr("line", "r0_ohm_per_km", expected_type="f8", default=np.nan) * multiplier
         pgm_lines["x0"] = self._get_pp_attr("line", "x0_ohm_per_km", expected_type="f8", default=np.nan) * multiplier
         pgm_lines["c0"] = c0_nf_per_km * length_km * parallel * 1e-9
-        pgm_lines["tan0"] = 0.0
```
mgovers, line 443: idem:

```py
invalid = np.isclose(c0_nf_per_km, 0.0)
pgm_lines["tan0"] = np.divide(
    self._get_pp_attr("line", "g0_us_per_km", expected_type="f8", default=0),
    c0_nf_per_km * (2 * np.pi * self.system_frequency * 1e-3),
    where=np.logical_not(invalid),
)
pgm_lines["tan0"][invalid] = 0.0
```

src/power_grid_model_io/converters/pandapower_converter.py
mgovers: nvm. just read https://github.com/PowerGridModel/power-grid-model-io/pull/326#issuecomment-3279663612 again. Resolving

> @mgovers @nitbharambe there is still a reliability issue on pypi.
I've checked that 2 No. lines causing this issue can be removed, the field is auto initialised with zero even when np.divide has some False entries in where.

src/power_grid_model_io/converters/pandapower_converter.py
mgovers: nvm. just read https://github.com/PowerGridModel/power-grid-model-io/pull/326#issuecomment-3279663612 again. Resolving

> @mgovers @nitbharambe there is still a reliability issue on pypi.
I've checked that 2 No. lines causing this issue can be removed, the field is auto initialised with zero even when np.divide has some False entries in where.

mgovers (approved):
Cfr. https://github.com/PowerGridModel/power-grid-model-io/pull/326#issuecomment-3279663612 , the original suggestion was perfect. I hereby approve

PR #329, opened 2025-09-17
Convert Name field to string type

Jerry-Jinfeng-Guo:
This PR converts the `Name` fields to string type to avoid scientific notations if the value is too big. Just a trial.

### DIFF
--- a/src/power_grid_model_io/data_stores/excel_file_store.py
+++ b/src/power_grid_model_io/data_stores/excel_file_store.py
@@ -93,9 +93,19 @@ class ExcelFileStore(BaseDataStore[TabularData]):
                 sheet_data = self._handle_duplicate_columns(data=sheet_data, sheet_name=xls_sheet_name)
                 sheet_data = self._process_uuid_columns(data=sheet_data, sheet_name=xls_sheet_name)
                 sheet_data = self._update_column_names(data=sheet_data)
+                # Only convert large integer values to strings for columns named 'Name'
+                for col in sheet_data.columns:
+                    if (col == "Name" or (isinstance(col, tuple) and col[0] == "Name")) and sheet_data[col].dtype in [
+                        "float64",
+                        "int64",
+                    ]:
+                        if (sheet_data[col].abs() >= 1e12).any():
+                            sheet_data[col] = sheet_data[col].apply(
+                                lambda x: str(int(x))
+                                if pd.notnull(x) and isinstance(x, (int, float)) and abs(x) >= 1e12
+                                else x
+                            )
                 return sheet_data
-
-            return sheet_loader
 
         data: Dict[str, LazyDataFrame] = {}
         for name, path in self._file_paths.items():


src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -88,14 +88,19 @@ def load(self) -> TabularData:
 
         def lazy_sheet_loader(xls_file: pd.ExcelFile, xls_sheet_name: str):
             def sheet_loader():
-                sheet_data = xls_file.parse(xls_sheet_name, header=self._header_rows)
+                preview = xls_file.parse(xls_sheet_name, header=self._header_rows, nrows=0)
+                columns = list(preview.columns)
+                dtype = {}
+                for col in columns:
+                    if col == "Name" or (isinstance(col, tuple) and col[0] == "Name"):
```
nitbharambe, line 95: Going generic for this case can be nice. For any column and not just Name.
(I do see this is a trial.)
If a generic way is explored and unsuitable then feel free to resolve and merge as well.

Jerry-Jinfeng-Guo: Good input, that would require some discussion as in the pipeline it is unclear what the users should expect as `dtype`.

tests/unit/data_stores/test_vision_excel_file_store.py
```diff
@@ -19,4 +21,56 @@ def test_header_rows(mock_excel_file: MagicMock):
...
+    mock_excel_file.return_value.sheet_names = ["test_sheet"]
+
+    preview_df = pd.DataFrame(columns=["Mock.Name", "Other.Column", "ID"])
+
+    def mock_parse(*args, **kwargs):
+        if kwargs.get("nrows") == 0:
+            return preview_df
+        else:
+            actual_data = {
+                "Mock.Name": [123456789000000, 987.654],
```
nitbharambe, line 40:
```suggestion
                "Mock.Name": [12345678900000000000, 987.654],
```

tests/unit/data_stores/test_vision_excel_file_store.py
```diff
@@ -19,4 +21,56 @@ def test_header_rows(mock_excel_file: MagicMock):
...
+
+    first_call = mock_excel_file.return_value.parse.call_args_list[0]
+    assert first_call[1]["nrows"] == 0
+
+    second_call = mock_excel_file.return_value.parse.call_args_list[1]
+    assert "dtype" in second_call[1]
+    assert "Mock.Name" in second_call[1]["dtype"]
+    assert second_call[1]["dtype"]["Mock.Name"] is str
+
+    assert result_df["Mock.Name"][0] == "123456789000000"  # Long int as string
```
nitbharambe, line 69:
```suggestion
    assert result_df["Mock.Name"][0] == "12345678900000000000"  # Long int as string
```

nitbharambe (approved):

PR #332, opened 2025-09-24
[FIX] Unbalance Percent in res_bus_3ph

furqan463:
To make it consistent with Pandapower. Currently Unbalance percent in pandapower_converter.py is V_zero / V_pos, while that defined in Pandapower is V_neg / V_pos. As per my knowledge, this is also consistent with IEC.

https://github.com/e2nIEE/pandapower/blob/develop/pandapower/results_bus.py#L71


mgovers:
Hi @furqan463, Good find!

@nitbharambe can you please review as the EE expert?

Can we somehow validate this? I do think this change is simple enough that we don't need explicit tests but it would be interesting to think about such a test case.

nitbharambe:
Good find indeed @furqan463 
Agreed per the IEC standard and the code at pandapower.

I think we were referring to pandapower documentation when we developed the feature which is likely incorrect:
image

nitbharambe (approved):

furqan463:
> I think we were referring to pandapower documentation

Documentation Now corrected for next release of PP.

https://github.com/e2nIEE/pandapower/pull/2726

Issue #338, opened 2025-10-10
[FEATURE] Graceful Handling of Missing Extra Columns in Vision Excel Files


juan-lopez-amezquita:
**Contex:**
Users can include additional metadata like `GUID` or `StationID` in the extra_info section via custom YAML mapping files. However, these extra columns may not always be present in the original Vision Excel file—especially when working with partial exports.

Currently, if an extra column such as `StationID` is specified in the mapping file but missing from the Excel file (e.g., in the Transformers or Three-winding transformers tables), the conversion fails.

**Proposal**
Update the Excel conversion logic to gracefully handle missing extra columns.

PR #344, opened 2025-11-26
Logic to handle optional_extra in Vision Excel converter input

Jerry-Jinfeng-Guo:
Closes https://github.com/PowerGridModel/power-grid-model-io/issues/338

In this PR:

- [x] Logic to handle `optional_extra`
- [x] Documentation
- [x] Test

### DIFF
--- a/src/power_grid_model_io/converters/tabular_converter.py
+++ b/src/power_grid_model_io/converters/tabular_converter.py
@@ -376,21 +376,19 @@ class TabularConverter(BaseConverter[TabularData]):
 
         # Collect all non-optional_extra column names
         regular_columns = set()
-        normalized_list = []
 
         for item in col_def:
             if isinstance(item, dict) and len(item) == 1 and "optional_extra" in item:
                 # This is an optional_extra section - we'll process it later
-                normalized_list.append(item)
+                pass
             else:
                 # This is a regular column
                 if isinstance(item, str):
                     regular_columns.add(item)
-                normalized_list.append(item)
 
         # Now process optional_extra sections and remove duplicates
         final_list = []
-        for item in normalized_list:
+        for item in col_def:
             if isinstance(item, dict) and len(item) == 1 and "optional_extra" in item:
                 optional_cols = item["optional_extra"]
                 if isinstance(optional_cols, list):
--- a/tests/unit/converters/test_tabular_converter.py
+++ b/tests/unit/converters/test_tabular_converter.py
@@ -1420,6 +1420,20 @@ def test_optional_extra__mixed_with_required(converter: TabularConverter):
     assert list(result["guid"]) == ["guid1", "guid2"]
 
 
+def test_optional_extra__mixed_with_required_missing_column(converter: TabularConverter):
+    """Test that missing required column raises error even with optional_extra present"""
+    # Arrange
+    data = TabularData(test_table=pd.DataFrame({"id": [1, 2], "name": ["node1", "node2"], "guid": ["guid1", "guid2"]}))
+    # "missing_col" is required but missing
+    col_def = ["name", "missing_col", {"optional_extra": ["guid", "station"]}]
+
+    # Act & Assert
+    with pytest.raises(KeyError, match="missing_col"):
+        converter._parse_col_def(
+            data=data, table="test_table", col_def=col_def, table_mask=None, extra_info=None, allow_missing=False
+        )
+
+
 def test_optional_extra__in_extra_info(converter: TabularConverter):
     """Test that optional_extra works correctly with _handle_extra_info"""
     # Arrange
@@ -1442,6 +1456,23 @@ def test_optional_extra__in_extra_info(converter: TabularConverter):
     assert extra_info[200]["guid"] == "guid2"
     assert "station" not in extra_info[100]
     assert "station" not in extra_info[200]
+
+
+def test_optional_extra__empty_list_with_extra_info(converter: TabularConverter):
+    """Test that empty optional_extra list results in no updates to extra_info"""
+    # Arrange
+    data = TabularData(test_table=pd.DataFrame({"id": [1, 2], "name": ["node1", "node2"], "guid": ["guid1", "guid2"]}))
+    uuids = np.array([100, 200])
+    extra_info: ExtraInfo = {}
+    col_def: dict[str, list[str]] = {"optional_extra": []}
+
+    # Act
+    converter._handle_extra_info(
+        data=data, table="test_table", col_def=col_def, uuids=uuids, table_mask=None, extra_info=extra_info
+    )
+
+    # Assert
+    assert len(extra_info) == 0
 
 
 def test_optional_extra__all_missing_no_extra_info(converter: TabularConverter):


src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -486,6 +503,14 @@ def _parse_col_def_column_name(
...
+            if allow_missing:
+                # Return empty DataFrame with correct number of rows when column is optional and missing
+                self._log.debug(
+                    "Optional column not found",
+                    table=table,
+                    columns=" or ".join(f"'{col_name}'" for col_name in columns),
+                )
+                return pd.DataFrame(index=table_data.index)
             columns_str = " and ".join(f"'{col_name}'" for col_name in columns)
             raise KeyError(f"Could not find column {columns_str} on table '{table}'")
```
mgovers:
```suggestion
            raise KeyError(f"Could not find column {columns_str} on table '{table}'") from e
```

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -404,15 +405,27 @@ def _parse_col_def(  # pylint: disable = too-many-arguments,too-many-positional-
           table: str:
           col_def: Any:
           extra_info: Optional[ExtraInfo]:
+          allow_missing: bool: If True, missing columns will return empty DataFrame instead of raising KeyError
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model-io/pull/344#discussion_r2576235175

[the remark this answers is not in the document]
mgovers: if i understand correctly, the only reason we need `allow_missing` is so that we can sanitize IF there are optional attributes in the result... That's a separate code path.

However, why can't we just gather all attributes and return empty dataframes by default, and then sanitize that a required attribute can't be an empty dataframe (unless of course there are no items in the first place). That way, we do not introduce a separate code path that is only triggered if there are optional attributes. That simplifies the logic a lot.

src/power_grid_model_io/converters/tabular_converter.py
```diff
@@ -484,10 +554,17 @@ def _parse_col_def_column_name(
...
+                # Return empty DataFrame with correct number of rows when column is optional and missing
+                self._log.debug(
+                    "Optional column not found",
+                    table=table,
+                    columns=" or ".join(f"'{col_name}'" for col_name in columns),
+                )
+                return pd.DataFrame(index=table_data.index)
             columns_str = " and ".join(f"'{col_name}'" for col_name in columns)
-            raise KeyError(f"Could not find column {columns_str} on table '{table}'")
+            raise KeyError(f"Could not find column {columns_str} on table '{table}'") from e
```
mgovers: I did some thinking on the original code here. It now understand why it did not have `raise from e` here: 

* either it is a string that can be converted to a float (e.g. `inf`; see the description in the `try:` statement), which should be interpreted as a float
* or is a column name which is absent which
  * either is allowed (new code)
  * or it is disallowed and should raise a `KeyError`

However, I dislike that original design choice. Instead, I am of the opinion that it should be something like

```py
if const_value := float_or_none(col_def):
    return self._parse_col_def_const(..., const_value=const_value)

# now, we know for sure that it is a column name that is absent
if allow_missing:
    return ...

raise KeyError(...)
```

where `float_or_none` can indeed be implemented as
```py
def float_or_none(value: str) -> float | None:
    try:
        return float(value)
    except ValueError:
        return None
```

The reason I **do not** want to raise in the except block is that the conversion to `float` is actually already the fall-back. In pseudo-code, the logic is actually something like

```py
if the column header is a regular column name:
    return column header as regular column name
elif the column header is a floating point value:
    return the column header as a floating point value
elif this specific column is optional
    return None
else:
    raise a KeyError because it was mandatory but is is not found
```

this also is an indication as to why the `raise ... from e` was disabled: the key-error is not the fall-back but the default - the conversion to float is actually the fall-back logic. But, again, there's a reason why `raise ... from e` is good practice: it shows the intention. In this original code, the intention was not clear.

tests/unit/converters/test_tabular_converter.py
```diff
@@ -1696,3 +1681,326 @@ def test_convert_col_def_to_attribute__pgm_data_without_dtype_names():
...
+    result = converter._parse_col_def(
+        data=data,
+        table="test_table",
+        col_def=["missing_col1", "missing_col2"],
+        table_mask=None,
+        extra_info=None,
+        allow_missing=True,
+    )
+    assert len(result.columns) == 0
+    assert len(result) == 3  # Should have same number of rows as original table
```
mgovers: and same contents

Jerry-Jinfeng-Guo: I don't understand what you mean.

src/power_grid_model_io/converters/tabular_converter.py
Jerry-Jinfeng-Guo: Rerverted.

mgovers: Done in #347

tests/unit/converters/test_tabular_converter.py
```diff
@@ -1696,3 +1681,326 @@ def test_convert_col_def_to_attribute__pgm_data_without_dtype_names():
...
+
+    # Test 9: Dict (optional_extra) - should automatically set allow_missing=True internally
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: The test specifies the label to `False` but is internally set to `True`, which is what this test tests.

tests/unit/converters/test_tabular_converter.py
```diff
@@ -1356,3 +1347,731 @@ def bool_fn_filter(row: pd.Series, **kwargs):
...
+    assert len(result) == 2
+    assert len(result.columns) == 0
+
+
+def test_optional_extra__mixed_with_required(converter: TabularConverter):
+    """Test mixing required and optional extra columns"""
+    # Arrange
+    data = TabularData(test_table=pd.DataFrame({"id": [1, 2], "name": ["node1", "node2"], "guid": ["guid1", "guid2"]}))
+    # Mix required columns with optional_extra
+    col_def = ["name", {"optional_extra": ["guid", "station"]}]
```
figueroa1395: Nitpick: I believe, correct me if I'm wrong, that there is not test that verifies that given this same configuration, if we have another column that is required then the parsing should raise. This is probably tested above somewhere in isolation, but not in combination with the optional stuff. 

I mean something like
```
col_def = ["name", "whatever", {"optional_extra": ["guid", "station"]}]

# check raises because "whatever" isn't present when parsing
```

this in the context of the same setup of this test.

tests/unit/converters/test_tabular_converter.py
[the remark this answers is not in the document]
figueroa1395: Again, I may lack context, but in the test case it is set to `False` and the expectation is that it is set to `True` when the correct logic is triggered, right? This means that the `missing_col` is ignored, so we only have one resulting column corresponding to the one that is present, which can only happen because `allow_missing` is now `True`, correct?

PR #346, opened 2025-12-09
Add packaging dependency

nitbharambe:
Added at #336

We should probably create a new action to run pipeline without dev dependencies. The dev dependencies here can actually create a very different set of dependencies compared to non group one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ requires-python = ">=3.11"
 dependencies = [
     "numpy>=2.0",
     "openpyxl",
+    "packaging>=25.0",
     "pandas",
     "power_grid_model>=1.8",
     "pyyaml",


mgovers:
> I believe the failing CI corresponds to a regression that was already [fixed](https://github.com/PowerGridModel/power-grid-model/pull/1213) in the main PGM repo. We'll have to wait for the update here.

No this is the Python 3.14.1 bug on `dataclass`. If you look at the final couple lines in the stacktrace:

```
 /Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/networkx/utils/__init__.py:7: in <module>
    from networkx.utils.configs import *
/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/networkx/utils/configs.py:8: in <module>
    @dataclass(init=False, eq=False, slots=True, kw_only=True, match_args=False)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/dataclasses.py:1426: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/dataclasses.py:1234: in _process_class
    cls = _add_slots(cls, frozen, weakref_slot, fields)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/dataclasses.py:1401: in _add_slots
    init_annotate = newcls.__init__.__annotate__
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: 'wrapper_descriptor' object has no attribute '__annotate__'. Did you mean: '__getstate__'?
```

PR #348, opened 2025-12-11
Enable UV in build and CI

mgovers:
PGM-IO equivalent of https://github.com/PowerGridModel/power-grid-model/pull/1206

Enables uv in build and CI.
Also homogenizes pre-commit, markdownlint and enables `mypy`

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1400,6 +1400,7 @@ source = { editable = "." }
 dependencies = [
     { name = "numpy" },
     { name = "openpyxl" },
+    { name = "packaging" },
     { name = "pandas" },
     { name = "power-grid-model" },
     { name = "pyyaml" },
@@ -1454,6 +1455,7 @@ test = [
 requires-dist = [
     { name = "numpy", specifier = ">=2.0" },
     { name = "openpyxl" },
+    { name = "packaging", specifier = ">=25.0" },
     { name = "pandas" },
     { name = "power-grid-model", specifier = ">=1.8" },
     { name = "pyyaml" },


.github/workflows/build-test-release.yml
```diff
@@ -89,11 +82,14 @@ jobs:
...
+      - name: Create environment
+        run: uv venv
+
       - name: Install built wheel file
-        run: pip install power-grid-model-io[dev]==${{ needs.build-python.outputs.version }} --find-links=wheelhouse
+        run: uv pip install power-grid-model-io==${{ needs.build-python.outputs.version }} --find-links=wheelhouse
 
       - name: Unit test and coverage
-        run: pytest --verbose
+        run: uv run --only-group test --no-editable pytest --verbose
```
TonyXiang8787: How will this work? So it does not force `uv` to install the source project into the `venv`?

.github/workflows/build-test-release.yml
figueroa1395: Out of scope of this PR, but shouldn't we now deprecate `3.11`? We seem a bit late there. Or at least keep it in mind until the fixed version of `3.14` is out.

.github/workflows/build-test-release.yml
figueroa1395: I like it this way. It's verbose in a good way. Shows intention. I also have no problem with the 2-stage logic.

pyproject.toml
```diff
@@ -38,22 +38,33 @@ dependencies = [
...
+    "mypy",
+    "numpy", # needed for mypy
+    "types-pyyaml", # needed for mypy
+    "types-requests", # needed for mypy
+]
+
+dev = [{ include-group = "test" }, { include-group = "lint" }]
+
 examples = [
     "power-grid-model>1.9.80",
```
figueroa1395: Does this also hold still?

pyproject.toml
```diff
@@ -38,22 +38,33 @@ dependencies = [
...
     "pandapower>2.11.1",
```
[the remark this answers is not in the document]
figueroa1395: I'll make time for it and make a separate PR then

.github/workflows/build-test-release.yml
figueroa1395: https://github.com/PowerGridModel/power-grid-model-io/issues/353

TonyXiang8787 (changes requested):
Please use the GitHub Apps setup as in here: 

https://github.com/PowerGridModel/power-grid-model/pull/1233

README.md
```diff
@@ -24,28 +24,31 @@ SPDX-License-Identifier: MPL-2.0
...
 ### Install from PyPI
 
 You can directly install the package from PyPI.
+Although it is opt-in, it is recommended to use [`uv`](https://github.com/astral-sh/uv) as
+your development environment manager.
 
-```
-pip install power-grid-model-io
+```sh
+uv add power-grid-model-io
```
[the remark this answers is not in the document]
figueroa1395: Given Tony's input I've switched the recommended commands to `uv pip install` as it replicates best what was there before and can be used ad hoc. If part of a project, users should know how this works better. Also, if they don't use `uv` but `pip`, it's easier to know what to use.

Issue #368, opened 2026-02-10
Breaking unit test in power_grid_model >=1.13.3


furqan463:
`power_grid_model` 1.13.3 introduced changes in component definition, which resulted failure of following unit test.
pytest tests/unit/converters/test_tabular_converter.py::test_convert_col_def_to_attribute

as `ComponentType.node` is now "node" instead of "ComponentType.node". resulting in regex pattern match failure.

PR #369, opened 2026-02-10
Fix breaking unit test with power-grid-model >= 1.13.3

furqan463:
**Fixes issue**: #368

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1355,19 +1355,19 @@ wheels = [
 
 [[package]]
 name = "power-grid-model"
-version = "1.13.4"
+version = "1.13.6"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "numpy" },
 ]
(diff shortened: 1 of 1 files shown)


tests/unit/converters/test_tabular_converter.py
```diff
@@ -275,10 +280,20 @@ def test_convert_col_def_to_attribute(
...
+        match_string = (
+            r"DataFrame for node.u_rated should contain a single column "
+            r"\(Index\(\['id_number', 'u_nom'\], dtype='(str|object)'\)\)"
+        )
+
     with pytest.raises(
         ValueError,
-        match=r"DataFrame for ComponentType.node.u_rated should contain a single column "
-        r"\(Index\(\['id_number', 'u_nom'\], dtype='(str|object)'\)\)",
+        match=match_string,
```
mgovers: I'm thinking whether this is the right solution. It's just a simple regex and the impact isn't huge, so you can also do something along the lines of this:

```suggestion
        match=r"DataFrame for (ComponentType\.)?node\.u_rated should contain a single column "
        r"\(Index\(\['id_number', 'u_nom'\], dtype='(str|object)'\)\)",
```

mgovers (approved):
@figueroa1395 and I were talking about this yesterday, whether the change in error message is actually desired. If I look at this specific error message, then the answer is yes so thanks for doing an attempt at this.

tests/unit/converters/test_tabular_converter.py
furqan463: Yes this makes more sense.

Issue #352, opened 2025-12-16
[IMPROVEMENT] Update `ruff` rules


figueroa1395:
**Describe the problem**

The [list](https://github.com/PowerGridModel/power-grid-model-io/blob/0b6a74304fe2cc9ec50f5eb22897fbf2a86fb100/pyproject.toml#L105) of `ruff` rules/checks is outdated in comparison to [PGM](https://github.com/PowerGridModel/power-grid-model/blob/b5348836d7a29dd45d047d2e14c488e5f385f005/pyproject.toml#L132).

**Improvement proposal**

Update the [list](https://github.com/PowerGridModel/power-grid-model-io/blob/0b6a74304fe2cc9ec50f5eb22897fbf2a86fb100/pyproject.toml#L105) of rules/check in `pyproject.toml` used by `ruff` in this repo to be aligned with the ones used in [PGM](https://github.com/PowerGridModel/power-grid-model/blob/b5348836d7a29dd45d047d2e14c488e5f385f005/pyproject.toml#L132). This will lead to many code changes that are to be fixed as well.

***Suggested steps***
1. Add a new rule not present in this [repo](https://github.com/PowerGridModel/power-grid-model-io/blob/0b6a74304fe2cc9ec50f5eb22897fbf2a86fb100/pyproject.toml#L105) but present in [PGM](https://github.com/PowerGridModel/power-grid-model/blob/b5348836d7a29dd45d047d2e14c488e5f385f005/pyproject.toml#L132) to the `pyproject.toml`.
2. Run `uv run ruff check`.
3. Fix the reported errors/warning by `ruff`.
4. Repeat until all the rules have been added.

PR #375, opened 2026-02-19
Update Ruff Rules (PT)

furqan463:
**Relates issue**: #352 

Add ruff lint rule pytest "PT".


figueroa1395 (approved):
@furqan463 Great job!

PR #386, opened 2026-03-02
[Feature] Asym Line support for Pandapower using PGM solver

### LINKED ISSUE #380: [FEATURE] Support AsymLine in PandaPowerConverter
Currently `pandapower` do not support `AsymLine`, however, it would be nice to add an options to convert `pandapower` lines as `AsymLine` in `PandaPowerConverter`.
We can do this by allowing user to redefine `std_types['line']` with `AsymLine` parameters within `PGM-IO`as an additional optional input to `PandaPowerConverter`. This will allow simulation of `AsymLine` with existing `pandapower` networks. Additionally, Since `Pandapower` allows additional user defined elements within `PandaPowerNet`, the users may define an extra `std_types['asym_line']` in `pandapowerNet` which will not be used within `pandapower` but will be used when converting to `PGM`. I'm also interested to implement this feature if deemed appropriate.

furqan463:
Fixes #380 

- Add an additional parameter `asym_line_params` to the `PandaPowerConverter`
- Checks if `asym_line` std type is present in `pp_input`. `asym_line_params` takes precedence in case the line parameters are present in both.
- For those lines where `asym_line` params are not defined, gets converted as `ComponentType.line` of `PGM`
- For those lines where `asym_line` params are defined, get converted as `ComponentType.asym_line` of `PGM`

### Checks

- [x] Implement output conversion for `asym_line`
- [x] Fix unit tests for `line`
- [ ] Add unit tests for `asym_line` input and output conversion and auxiliary methods
- [ ] Fix test coverage

### DIFF
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -2057,6 +2057,8 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         pp_output_lines_3ph = None
 
         for line_type in [ComponentType.line, ComponentType.asym_line]:
+            if (line_type not in self.pgm_input_data) or (line_type not in self.pgm_output_data):
+                continue
             pgm_input_lines = self.pgm_input_data[line_type]
             pgm_output_lines = self.pgm_output_data[line_type]
             if (pgm_input_lines.size == 0) or (pgm_output_lines.size == 0):


src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2638,13 +2757,20 @@ def _get_pp_attr(
         attr_data = pp_component_data[attribute]
 
         # If any of the attribute values are missing, and a default is supplied, fill the nans with the default value
-        nan_values = np.equal(attr_data, None) if attr_data.dtype is np.dtype("O") else np.isnan(attr_data)  # type: ignore
+        nan_values = attr_data.isna()  # type: ignore
```
furqan463: The existing check failed with Pandas 3.0, with default StringDtype.
Although `pandapower` currently forces dtype `object`, the change is cleaner and will work with future changes in `pandapower` when it supports `pandas>=3.0`, since attr_data is always `pd.DataFame`

PR #377, opened 2026-02-24
Code quality: update Ruff Rules (Bandit, S)

furqan463:
**Relates issue**: #352 

Adds ruff lint rules Bandit "S".

PR #387, opened 2026-03-05
Revert to default XML parser

figueroa1395:
In #377 we switched from `xml` to `defusedxml` because of `ruff`'s [suggestion](https://docs.astral.sh/ruff/rules/suspicious-xml-expat-builder-usage/). This suggestion is outdated and only applies to `python < 3.11`. The reason is that the vulnerabilities existed only for `python < 3.11`,  as`xml` was built with `expat = 2.6.0` which was the source of such vulnerabilities. However, this no longer holds as `expat = 2.7.1` is used for `python >= 3.11` and `python <= 3.15`, removing such vulnerabilities.

See the following for references on the topic:
- https://github.com/python/cpython/pull/135294
- https://github.com/python/cpython/issues/127502
- https://docs.python.org/3/library/xml.html#the-defusedxml-package

In addition, the reason I found this is because the [feedstock build](https://github.com/conda-forge/power-grid-model-io-feedstock/pull/128#issuecomment-3998829039) failed due to `defusedxml` not being a "default" python dependency, which made me question the change and dig deeper into the issue.

C.C. @furqan463 I thought you would like to be tagged on this one.

### DIFF
--- a/docs/converters/tabular_converter.md
+++ b/docs/converters/tabular_converter.md
@@ -381,9 +381,12 @@ Mapping files enable the specification of custom mappings or filter functions.
 These functions can come from the `power-grid-model-io` library, be user-provided, or even supplied by third parties.
 To ensure security, we have implemented several measures.
 Best practices are recommended to prevent malicious code execution.
-XML parsing is performed using the defusedxml library instead of the standard library xml module.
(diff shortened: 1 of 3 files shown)


mgovers (approved):
unfortunate that ruff is not updated. good find

PR #393, opened 2026-03-16
Print network to PDF

Jerry-Jinfeng-Guo:
Functionality to print a PGM network data to a PDF file with electrical diagrams.

- [x] function
- [x] test
- [x] doc/example


mgovers (commented):
Very cool but is pgm-io the right place for this? Shouldn't this be part of pgm-ds?

mgovers:
Closing this PR as per the offline discussion. It can be re-opened once relevant again.

PR #406, opened 2026-04-01
Pin tests to deepdiff~=8.6

figueroa1395:
Fixes https://github.com/conda-forge/power-grid-model-io-feedstock/pull/143.

`pandapower=3.2.1` has `deepdiff~=8.6` dependency. Hence in this PR we pin to this version for the tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ test = [
     "pydantic>2", # Used in unit tests
     "pandapower>2.11.1",
     "lxml", # missing pandapower dependency
+    "deepdiff~=8.6" # pandapower 3.2.1 has requirement deepdiff~=8.6
 ]
 
 lint = [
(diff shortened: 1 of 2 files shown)


mgovers (approved):

pyproject.toml
```diff
@@ -47,6 +47,7 @@ test = [
     "pydantic>2", # Used in unit tests
     "pandapower>2.11.1",
     "lxml", # missing pandapower dependency
+    "deepdiff~=8.6" # pandapower 3.2.1 has requirement deepdiff~=8.6
```
mgovers, line 50: please do this change directly to the other PRs. We have mainainer rights so that should work. Otherwise, you'd have to bump the build number manually

mgovers (changes requested):

pyproject.toml
figueroa1395: What do you mean? As in do it directly in the feedstock? If we merge this PR with the fix here and this PR is merged, that would close the other unmerged PRs, and that should be okay, shouldn't it?

mgovers: no you have to do it on one of the already open PRs. That would not close them unless the build succeeds and they are merged

furqan463, line 50: Doesn't `uv` handle this automatically? It'll always install `deepdiff~=8.6` when `pandapower` is included in the dependency. Do we need `deepdiff` directly in `pgm_io`?

furqan463: `pandapower` has restriction on `pandas` as well. But `pandas 3.0` does not get installed in https://github.com/conda-forge/power-grid-model-io-feedstock/pull/143

figueroa1395: @furqan463 what you say makes sense. I'll investigate this a bit more and try to figure out why this isn't the case for `deepdiff`. Thanks for pointing it out. I'll write my conclusions in this PR.

figueroa1395: @mgovers doing it in https://github.com/conda-forge/power-grid-model-io-feedstock/pull/143/

figueroa1395: So `feedstock` doesn't use `uv` yet, so that's the reason is not a problem for us directly, but it's a problem there. I'm directly fixing on their side in this [PR](https://github.com/conda-forge/power-grid-model-io-feedstock/pull/143/), so then is retroactive for the following one in the queue.

I'll close this PR now and track it there.

figueroa1395:
Closing because the fix is happening directly at the feedstock: https://github.com/conda-forge/power-grid-model-io-feedstock/pull/143/.

Issue #315, opened 2025-06-16
[IMPROVEMENT] `Enum`s for "external" components


figueroa1395:
As pointed out in https://github.com/PowerGridModel/power-grid-model-io/pull/295#discussion_r2149205652, some of the component type names in other libraries, such as in `pandapower`, are different to the names used in the `power-grid-model`. For this reason, it would be good practice to create `Enum`s, similar to what was done in https://github.com/PowerGridModel/power-grid-model/pull/648 for the `power-grid-model` component types and  implemented in `power-grid-model-io` in https://github.com/PowerGridModel/power-grid-model-io/pull/295, to keep track of the component types used in the libraries we convert from/to, as well as to keep reducing raw `str`s.

This may not directly affect the user, but will definitely keep the code implementation cleaner, and also help keeping track of what elements come from other libraries we convert to/from.

furqan463:
Regarding reduction in using raw `str`, `AttributeType` has just been implemented in `power_grid_model`. But how can we implement `AttributeType` in this repo without restricting the `power_grid_model` to latest version that implements `AttributeType`.

mgovers:
Another minor addtion: let's say that PGM-IO now starts adding conversion for a new PGM parameter that PP already supported before, both PGM-IO and PGM need to be version bumped anyways

mgovers:
I'd consider pandapower attributes/tables/... are code-internal. As long as you prefix them `_` (i.e., make them module-private), any name goes. If we decide to change or split it up in the future, that's perfectly possible, because it doesn't affect the API. Just go with your gut 😄

PR #407, opened 2026-04-01
[IMPROVEMENT] `Enum` for `pandapower` component tables

furqan463:
Relates Issue #315 

Introduces `Enum` for `pandapower` component tables and result tables.

- `Enum` for attributes will be introduced in followup PRs.
- Examples not being updated as the enums are created as private https://github.com/PowerGridModel/power-grid-model-io/issues/315#issuecomment-4168233197 and not part of API.


src/power_grid_model_io/_enum.py
mgovers, line 1: I don't think it's good to create a separate `enum.py` module because it's visible to users.

I see 2 options:

* [ ] create the `_PandapowerTable` enum in the `pandapower_converter.py`, as it's very specific to that module.
* [ ] if you really want to separate the enum declaration from the usage, you can also create `_enum.py` instead. This will make the entire enum module private. You can then create `PandapowerTable` (notice the missing leading `_`; see my other comment). We also use this trick in the PGM itself.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -30,6 +30,7 @@
 
 from power_grid_model_io.converters.base_converter import BaseConverter
 from power_grid_model_io.data_types import ExtraInfo
+from power_grid_model_io.enum import _PandapowerTable as PpTable
```
mgovers:
```suggestion
from power_grid_model_io.enum import _PandapowerTable as _PpTable
```

`PpTable` is not to be exposed to the user, so it should be marked module-private with a leading `_`

src/power_grid_model_io/enum.py
```diff
@@ -0,0 +1,65 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+from enum import StrEnum
+
+
+class _PandapowerTable(StrEnum):
```
mgovers: `from my_lib.foo import _bar` is generally bad practice. prefix `_` means the object is intended to be module-private, so it indicates that `_bar` is not to be used outside `bar`

On the other hand, `from my_lib._foo import bar` indicates that `_foo` is library-internal, i.e., only to be used in the implementation of `my_lib`. To prevent accidentally exposing library-internal code where internal stuff is imported, you can use a pattern like `from my_lib._foo import bar as _bar`. Since exposing to the user is a one-way path, it's important to spend extra care on handling this correctly.

Note: `from other_lib._baz import bar` would be bad practice, as `_baz` is private to the implementation of `other_lib`, which is not the library we're currently working on.

mgovers (changes requested):
most of it looks good but i'm clicking request changes because there's an API change that should not be in there. It needs to be double-checked that that is resolved.

src/power_grid_model_io/_enum.py
furqan463: Actually I thought we may decide to add some enums in API as well, so I added enum.py. however, I'll make it _enum.py for now.

furqan463: Done.

src/power_grid_model_io/converters/pandapower_converter.py
furqan463: Done.

src/power_grid_model_io/enum.py
furqan463: Done.

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -23,12 +23,12 @@
     initialize_array,
 )
 
+from power_grid_model_io._enum import PandapowerTable as _PpTable
```
mgovers, line 26: in tests, you don't necessarily need to do the leading underscore alias because they are not published, but good that you thought about it 👍 it's fine to leave as-is

tests/unit/converters/test_pandapower_converter_input.py
```diff
@@ -831,44 +833,44 @@ def test_create_pgm_input_sym_loads__bad_input():
...
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "p_a_mw", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "p_b_mw", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "p_c_mw", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "q_a_mvar", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "q_b_mvar", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "q_c_mvar", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "scaling", expected_type="f8")
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "in_service", expected_type="bool", default=True)
-    converter._get_pp_attr.assert_any_call(PpTable.asymmetric_load, "type", expected_type="O", default=None)
+    converter._get_pp_attr.assert_any_call(_PpTable.asymmetric_load, _PpTable.bus, expected_type="u4")
```
mgovers: here and elsewhere is it intended that `bus` here is `PpTable` and not an attribute?

mgovers (dismissed):

tests/unit/converters/test_pandapower_converter_input.py
furqan463: Yes it's a mistake, I'll fix it.

furqan463: Fixed.

mgovers (approved):

PR #409, opened 2026-04-09
Pandapower attr enum

### LINKED ISSUE #315: [IMPROVEMENT] `Enum`s for "external" components

furqan463:
Fixes #315 
as enums are only required for `pandapower`. `json`, `excel` and `vision` converters do not have raw strings in src files.

Adds attribute enums for `pandapower`.


furqan463:
Created attr enums by iterating over an empty `pandapower` network tables and adding all columns to a `set` then had to enter some manually. However, there are several attr enums that are not being used inside `pgm_io`. Should we retain the exhaustive enum as it is, are somehow retain only those attributes that are actually being utilized in `pgm_io`?.

figueroa1395:
Hello @furqan463,

Thanks for the contribution. At the moment we are a bit occupied, so we expect to review and give feedback sometime next week.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -1868,29 +1916,33 @@ def join_currents(table: str, bus_name: str, i_name: str) -> pd.DataFrame:
             for bus_name, i_name in attr_names.items()
             if not rest_switches_absent[table]
         ]
-        all_i_df = pd.DataFrame(columns=[_PpTable.bus, "element", "et", "i_ka"]) if not dfs else pd.concat(dfs)
+        all_i_df = (
+            pd.DataFrame(columns=[_PpTable.bus, _PpAttr.element, _PpAttr.et, _PpAttr.i_ka])
```
mgovers: is it correct that this `_PpTable.bus` and not `_PpAttr.bus`? same for the following lines

NOTE: should not be a breaking change to users:

```py
>>> from enum import Enum
>>> from pandas import DataFrame as DF
>>>
>>> from enum import StrEnum 
>>> class Foo(StrEnum):
...   a = "a"
... 
>>> class Bar(StrEnum):
...   a = "a"
... 
>>> df = DF(columns=[Foo.a])
>>> df[Foo.a]
Series([], Name: a, dtype: object)
>>> df[Bar.a]
Series([], Name: a, dtype: object)
```

mgovers (dismissed):
One minor question, all the reset LGTM.

src/power_grid_model_io/converters/pandapower_converter.py
furqan463: Yes this should be `_PpAttr.bus`.

furqan463: Reviewed the whole code again for "bus" as this is common in `_PpTable` and `_PpAttr` and fixed all incorrect references.

mgovers (approved):

PR #427, opened 2026-05-18
Pandapower converter: allow string dtype columns in the pp converter

### LINKED ISSUE #422: [FEATURE] handle `pd.StringDtype` columns in pandapower_converter
### Describe the feature

With the next major version of pandapower (release/v4.0.0 branch) there will be issues with the pandapower_converter due to pandapower now using `pd.NA` for unset fields thus using `pd.StringDtype` for `name`/`type` etc. columns.

Since the pandapower_converter uses `np.isnan` to check for nan_values it fails as these columns cannot be handled.

### Code snippet
Error is in `power_grid_model_io.converter.pandapower_converter.py` on Line 2641
```python
nan_values = np.equal(attr_data, None) if attr_data.dtype is np.dtype("O") else np.isnan(attr_data)
```
This should most likely use `pd.isna(attr_data)` instead of `np.isnan`

### To Reproduce

```python
from pandapower.run import runpp_pgm
from pandapower.networks import example_simple
import warnings

warnings.simplefilter(action='ignore', category=FutureWarning)

net = example_simple()
# Remove Generator since its not supported yet
net["gen"] = net["gen"].iloc[:0]

# Convert columns that cannot be handled by power_grid_models for success
# strings = {
#     "bus": ["name", "type"],
#     "load": ["name", "type"],
#     "sgen": ["name"],
#     "switch": ["type"],
#     "shunt": ["name"],
#     "line": ["name", "std_type", "type"],
#     "trafo": ["name", "std_type", "vector_group", "tap_side"],
# }
# for table, columns in strings.items():
#     for col in columns:
#         net[table][col] = net[table][col].astype(object)

runpp_pgm(net)
```

### Expected behavior

Well, not getting an error for a start. ;-)

### Additional context

https://github.com/e2nIEE/pandapower/issues/2910

The issue opened for pandapower, it is not as detailed, as any mention of pd.StringDtype being a problem is missing there, I will update it soon.

### Community Meeting

- [ ] I wish to discuss this bug in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

mgovers:
Fixes #422 

This is relevant for pandapower 4.x compatibility

### DIFF
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -47,6 +47,7 @@ except PackageNotFoundError:
     PP_CONVERSION_VERSION = PP_COMPATIBILITY_VERSION_3_4_0  # assume latest compatible version by default
 
 _NOT_SET_STR = "__PGM_PP_STR_NOT_SET"
+_PD_NA_TYPE = type(pd.NA)
 
 
 def get_loss_params_3ph():
@@ -2784,7 +2785,7 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         """
 
         @lru_cache
-        def vector_group_to_winding_types(vector_group: str | None | pd.missing.NAType) -> pd.Series:
+        def vector_group_to_winding_types(vector_group: str | None | _PD_NA_TYPE) -> pd.Series:  # type: ignore[valid-type]
             if pd.isna(vector_group) or vector_group is None or vector_group == _NOT_SET_STR:
                 return pd.Series([np.nan, np.nan])
 
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -47,7 +47,6 @@ except PackageNotFoundError:
     PP_CONVERSION_VERSION = PP_COMPATIBILITY_VERSION_3_4_0  # assume latest compatible version by default
 
 _NOT_SET_STR = "__PGM_PP_STR_NOT_SET"
-_PD_NA_TYPE = type(pd.NA)
 
 
 def get_loss_params_3ph():
@@ -2785,7 +2784,7 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
         """
 
         @lru_cache
-        def vector_group_to_winding_types(vector_group: str | None | _PD_NA_TYPE) -> pd.Series:  # type: ignore[valid-type]
+        def vector_group_to_winding_types(vector_group: str | None | pd.missing.NAType) -> pd.Series:
             if pd.isna(vector_group) or vector_group is None or vector_group == _NOT_SET_STR:
                 return pd.Series([np.nan, np.nan])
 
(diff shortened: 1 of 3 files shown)


figueroa1395 (dismissed):

tests/validation/converters/test_pandapower_converter_input.py
```diff
@@ -175,6 +175,15 @@ def test_simple_example():
     _, _ = pp_converter.load_input_data(pp_net)
 
 
+@pytest.mark.filterwarnings("error")
+def test_simple_example_with_strings():
+    pp_net = example_simple()
+    pp_net["load"]["type"] = pp_net["load"]["type"].astype(pd.StringDtype())
+    pp_net[_PpTable.gen] = pp_net[_PpTable.gen].iloc[:0]
```
mgovers: as found by @nitbharambe , explicitly setting `pd.NA` also needs to be caught

mgovers: resolved

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -46,6 +46,9 @@
 except PackageNotFoundError:
     PP_CONVERSION_VERSION = PP_COMPATIBILITY_VERSION_3_4_0  # assume latest compatible version by default
 
+_NOT_SET_STR = "__PGM_PP_STR_NOT_SET"
+_PD_NA_TYPE = type(pd.NA)
```
figueroa1395, line 50: In summary, these are the two changes, right?
The first is the default, whereas the second is the new "nan". Also it's accompanied by a "new"` pd.isna`, right?

I'm checking if I understand the changes.

src/power_grid_model_io/converters/pandapower_converter.py
```diff
@@ -2860,7 +2868,11 @@ def _get_pp_attr(
         attr_data = pp_component_data[attribute]
 
         # If any of the attribute values are missing, and a default is supplied, fill the nans with the default value
-        nan_values = np.equal(attr_data, None) if attr_data.dtype is np.dtype("O") else np.isnan(attr_data)  # type: ignore
+        nan_values = (
+            np.equal(attr_data, None) | pd.isna(attr_data)  # type: ignore
```
figueroa1395, line 2872: It isn't clear to me why `| pd.isna(attr_data)`

src/power_grid_model_io/converters/pandapower_converter.py
mgovers: it's the other way around: `pd.NA` is added, but the existing workflow did not work due to default `None` parameters. To fix this, we had to add a dummy default fallback string because `pd.NA` seems to behave strangely when going from pandas to numpy (which makes sense because `pd.NA` does not abide by the "normal" structured array types) and `None` is illegal as a fallback in recent numpy versions for the same reasons. `pd.isna` catches all but no longer supports `None` as a default.

src/power_grid_model_io/converters/pandapower_converter.py
mgovers: The input type may be either `None` (older pandapower versions) or `pd.NA` (`pandapower>=4.0.0`), but `pd.isna` on older pandas versions does not catch `None` so removing `np.equal(attr_data, None)` would be a breaking change...

nitbharambe (approved):

PR #451, opened 2026-07-16
Bugfix: clean up excel file resources

### LINKED ISSUE #452: [BUG] Excel files are not closed
### Describe the bug

The pandas `ExcelFile` instances are lazy-loaded. That makes closing potentially ambiguous, due to unknown lifetimes.

### Input Data Validity

- [x] I checked the validity of input data using the helper function ([examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Validation%20Examples.html)).

### Data Dump

_No response_

### To Reproduce

See the validation cases in #451 . Note that the error raised is in the `pandapower` test cases, which seems weird at first because the issue is actually with the VisionExcelConverter, but it actually makes sense because the `pandapower` tests are coincidentally where the garbage collector decides to clean up resources, resulting in the unraisable exception.

### Expected behavior

the excel files are closed.

### Screenshots

_No response_

### Additional context

_No response_

### Community Meeting

- [ ] I wish to discuss this bug in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

mgovers:
by default, `pytest` does not run the validation cases. E.g. `uv run pytest` will return

```
800 passed in 8.77s
```

# this branch
```
1404 passed, 6 warnings in 9.87s
```

c.c. @furqan463 if you remember, we have talked about this. This enables validation pytest cases by default.

## Related bug

During creation of this PR, #452 was found. This branch also contains the fix for it.

Fixes #452

### DIFF
--- a/src/power_grid_model_io/data_stores/excel_file_store.py
+++ b/src/power_grid_model_io/data_stores/excel_file_store.py
@@ -102,7 +102,7 @@ class ExcelFileStore(BaseDataStore[TabularData]):
             return sheet_loader
 
         data: dict[str, LazyDataFrame] = {}
-        open_readers: pd.ExcelFile = []
+        open_readers: list[pd.ExcelFile] = []
         for name, path in self._file_paths.items():
             open_readers.append(pd.ExcelFile(path))
             excel_file = open_readers[-1]
--- a/src/power_grid_model_io/data_stores/excel_file_store.py
+++ b/src/power_grid_model_io/data_stores/excel_file_store.py
@@ -102,8 +102,10 @@ class ExcelFileStore(BaseDataStore[TabularData]):
             return sheet_loader
 
         data: dict[str, LazyDataFrame] = {}
+        open_readers: pd.ExcelFile = []
         for name, path in self._file_paths.items():
-            excel_file = pd.ExcelFile(path)
+            open_readers.append(pd.ExcelFile(path))
+            excel_file = open_readers[-1]
             for sheet_name in excel_file.sheet_names:
                 loader = lazy_sheet_loader(excel_file, sheet_name)
                 # If the Excel file is not the main file, prefix the sheet name with the file name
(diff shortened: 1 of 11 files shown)


nitbharambe (dismissed):

furqan463:
> c.c. @furqan463 if you remember, we have talked about this

I don't recall, but this is good.

figueroa1395 (dismissed):

src/power_grid_model_io/data_stores/excel_file_store.py
```diff
@@ -102,16 +102,19 @@ def sheet_loader():
             return sheet_loader
 
         data: dict[str, LazyDataFrame] = {}
+        open_readers: pd.ExcelFile = []
```
nitbharambe:
```suggestion
        open_readers: list[pd.ExcelFile] = []
```

tests/unit/converters/test_base_converter.py
```diff
@@ -135,6 +165,16 @@ def test_load_sc_output_data(converter: DummyConverter):
     assert data == {"foo": 1}
 
 
+def test_load_asym_output_data__resource_cleanup(converter: DummyConverter):
```
nitbharambe: We missed a `test_load_sc_output_data__resource_cleanup`

nitbharambe: @copilot add it. Similar to other tests

nitbharambe:
Rest LGTM

tests/unit/converters/test_base_converter.py
nitbharambe: Good to know:
> Note: the push to the branch is currently blocked by a repository branch protection rule (Cannot update this protected ref). The commit exists locally and will need the protection rule to be resolved before it can be pushed.

Lets discuss if we should restrict / allow it.

mgovers: done manually instead

nitbharambe (approved):

Issue #209, opened 2023-12-13
[FEATURE] PSS/E conversion to PGM


petersalemink95:
As requested during the 4th PGM meet-up

mgovers:
> I'm interested in contributing to this feature.

Hi @furqan463, Great to hear! Would you be able to make it for the community meeting in at 16:00 CET (+01:00)? Maybe we can find some time to discuss this there. You can register via https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745 if you have not done so already.

PR #464, opened 2026-08-05
[FEATURE] ADD PSS/E conversion to PGM

furqan463:
Relates to Issue #209 

### Changes proposed in this PR include

This PR adds converter for PSS/E RAWX file (json file) to PGM conversion.

### Checks

- [ ] Add unit tests
- [ ] Add validation tests


furqan463:
@mgovers Hi!, this is a work in progress, I created the PR to discuss important things along the way.

For now, Please review, I've moved basic structure of `tabular_data` to `tabular_data_base` so I could reuse existing code. If it's ok, then I can proceed further.

mgovers:
> @mgovers Hi!, this is a work in progress, I created the PR to discuss important things along the way.
> 
> For now, Please review, I've moved basic structure of `tabular_data` to `tabular_data_base` so I could reuse existing code. If it's ok, then I can proceed further.

Hi @furqan463,

Thank you for the valuable contribution! We are a bit occupied right now and are trying to finish a few things that have been dragging on for a bit longer than we originally intended. We will definitely make time this week but I am not sure when exactly. We'll keep you posted.

src/power_grid_model_io/data_stores/rawx_file_store.py
```diff
@@ -0,0 +1,63 @@
...
+        network = json_data["network"]
+
+        for component in network:
+            columns = network[component]["fields"]
+            component_data = network[component]["data"]
+            if len(component_data) == 0:
+                data[component] = pd.DataFrame(columns=columns)
+                continue
+            if not isinstance(component_data[0], list):
+                component_data = [component_data]
```
mgovers, line 58: what other types can it be? is the problem that the nesting is not deep enough (e.g. it's only a single row instead of a 2D table)?

can you make a named function of this

src/power_grid_model_io/converters/psse_converter.py
```diff
@@ -0,0 +1,241 @@
...
+        c_nf_per_km = self._get_pp_attr(_PpTable.line, _PpAttr.c_nf_per_km, expected_type="f8", default=0)
+        c0_nf_per_km = self._get_pp_attr(_PpTable.line, _PpAttr.c0_nf_per_km, expected_type="f8", default=0)
+        multiplier = length_km / parallel
+
+        pgm_lines = initialize_array(
+            data_type=DatasetType.input, component_type=ComponentType.line, shape=len(pp_lines)
+        )
+        pgm_lines[AttributeType.id] = self._generate_ids(_PpTable.line, pp_lines.index)
+        pgm_lines[AttributeType.from_node] = self._get_pgm_ids(
+            _PpTable.bus, self._get_pp_attr(_PpTable.line, _PpAttr.from_bus, expected_type="u4")
```
mgovers: does it make sense to have the function named `_get_pp_attr` even though it's not pandapower but pss/e? you may want to rename this function (please do so in a separate PR off `main`)

src/power_grid_model_io/converters/psse_converter.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        pgm_nodes[AttributeType.id] = self._generate_ids(_PsseTable.bus, busses.ibus)
+        pgm_nodes[AttributeType.u_rated] = self.psse_input_data.get_column(
+             table_name=_PsseTable.bus,
+             column_name="baskv"
+        ).to_numpy() * 1e3
+
+        self.pgm_input_data[ComponentType.node] = pgm_nodes
+
+    def _create_pgm_input_lines(self):
```
mgovers, line 117: i see quite a lot of things that are suspiciously similar to pandapower. how similar are the data formats? would it make sense to make separate functions that can be used by both converters?

mgovers (commented):
my first impression is that you're on the right track.
* Can you please also add some example data?
  * Or is that proprietary?
    * If so, how would we test it?
  * If it is open-source with a permissive license, then we can also add it to the project, even if it is not MPL, but LGPL/GPL are unfortunately not permissive enough.

Do you need an in-depth review already or do you have enough for the time

furqan463:
> Do you need an in-depth review already or do you have enough for the time

No No, I've just started working on it, I'll let you know when I need a review. For now I just wanted to confirm creating `tabular_data_base`.

src/power_grid_model_io/data_stores/rawx_file_store.py
furqan463: Yes, in case of single record, there's only simple list here instead of a nested list.

src/power_grid_model_io/converters/psse_converter.py
furqan463: This section is copied from `pandapower_converter`, I haven't touched it yet.

src/power_grid_model_io/converters/psse_converter.py
furqan463: I'm copying the structure from `pandapower_converter`, but overall it'll be different, specially after using unit conversion feature of `tabular_data_base`. We can come back to this point once it's finished.

furqan463:
> * Can you please also add some example data

Yes I'll add data in example and tests. The examples accompanied with PSSE software must be proprietary, I plan to create our own rawx file from one of the pp or pgm examples.

src/power_grid_model_io/data_stores/rawx_file_store.py
furqan463: > can you make a named function of this

sure.

PR #472, opened 2026-08-17
disabled markdownlint biome

Jerry-Jinfeng-Guo:


.pre-commit-config.yaml.jinja
```diff
@@ -34,8 +34,3 @@ repos:
         language: system
         pass_filenames: false
         always_run: true
-  - repo: https://github.com/igorshubovych/markdownlint-cli  # outside pypi
```
nitbharambe, line 37: The jinja file is redundant here as well.

Jerry-Jinfeng-Guo: See d3d7b9b

TonyXiang8787 (approved):

PR #482, opened 2026-08-25
Remove pip from dependabot, ignore pandas major version

nitbharambe:
Apparently uv ecosystem takes care of lockfile and also pyproject related updates. Only requirements.txt is additionally covered in pip compared to uv ecosystem. Hence pip one is redundant.

Dependency update ignored because of reason stated in the comment (and #480 )

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,7 @@ updates:
           - "*"
     ignore:
       # pandapower does not yet support pandas 3.x https://github.com/e2nIEE/pandapower/issues/2861
+      # Corresponding PGM-IO issue https://github.com/PowerGridModel/power-grid-model-io/issues/483
       - dependency-name: "pandas"
         update-types: ["version-update:semver-major"]
     cooldown:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,18 +16,6 @@ updates:
           - "*"
     cooldown:
       default-days: 7
-      
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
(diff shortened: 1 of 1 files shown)


mgovers (dismissed):

.github/dependabot.yml
```diff
@@ -38,5 +26,9 @@ updates:
       all-dependencies:
         patterns:
           - "*"
+    ignore:
+      # pandapower does not yet support pandas 3.x https://github.com/e2nIEE/pandapower/issues/2861
+      - dependency-name: "pandas"
+        update-types: ["version-update:semver-major"]
```
mgovers, line 33: optionally: also refer to the PGM-IO issue

nitbharambe: We dont have one. Do I create one? We dont have any additional tasks to do to support pandas 3.0.

mgovers: probably good to create one specifically targeting cleaning this exclusion up after the pandapower version bump

.github/dependabot.yml
```diff
@@ -38,5 +26,9 @@ updates:
       all-dependencies:
         patterns:
           - "*"
+    ignore:
+      # pandapower does not yet support pandas 3.x https://github.com/e2nIEE/pandapower/issues/2861
```
nitbharambe, line 30:
```suggestion
      # pandapower does not yet support pandas 3.x https://github.com/e2nIEE/pandapower/issues/2861
      # Corresponding PGM-IO issue https://github.com/PowerGridModel/power-grid-model-io/issues/483
```

mgovers (approved):

PR #490, opened 2026-09-06
Fix: Reliability issue raised by SonarQube

furqan463:
Hopefully reliability issue raised by SonarQube will be resolved by making the dataframe read explicit.

### DIFF
--- a/src/power_grid_model_io/converters/pandapower_converter.py
+++ b/src/power_grid_model_io/converters/pandapower_converter.py
@@ -1470,8 +1470,8 @@ class PandaPowerConverter(BaseConverter[PandaPowerData]):
 
                 # Now add the active and reactive powers to the pp busses
                 # Note that the units are incorrect; for efficiency, unit conversions will be applied at the end.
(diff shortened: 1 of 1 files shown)


mgovers (approved):

furqan463:
@mgovers can we run SonarQube scan here before merging?

furqan463:
> @mgovers can we run SonarQube scan here before merging?

BTW, what's the purpose of Sonar Cloud in workflow when SonarQube scan is skipped.
Shouldn't this whole workflow be made conditional with Sonar Cloud login access?

mgovers:
uhm, yes... good point...

So the reason is https://github.com/PowerGridModel/power-grid-model-io/actions/runs/34042967971/workflow?pr=490#L44 which explicitly states that PGM sonar cloud cannot run on a repo that is not owned by the PGM org. That restriction makes sense because it is a potential attack surface.

We will need to think about how to handle this. Let me quickly take this to the team. Most likely, a local Sonar Scanner execution is required for this.

Jerry-Jinfeng-Guo:
We will create a feature branch so that we can base this PR to a PGM org branch for the qube to run.

PR #497, opened 2026-09-10
Exclude Validation tests from unit tests action

furqan463:
In #451 validation tests were made to run by default with `pytest`, however, after this change validation tests are now included in the unit tests action. Because of this, validation tests are being executed twice and coverage is also being applied to the validation tests.

A failed validation test results in a failed unit test action.

This PR makes unit test action run unit tests explicitly with coverage, while validation tests will remain in validation test action without coverage.

### Checks

- [ ] If changes are related to CI/CD, are they verified via a manual run on this branch?

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -82,7 +82,7 @@ jobs:
         run: uv pip install power-grid-model-io==${{ needs.build-python.outputs.version }} --find-links=wheelhouse
 
       - name: Unit test and coverage
-        run: uv run --frozen --only-group test --no-editable pytest --verbose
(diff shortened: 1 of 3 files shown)


furqan463:
pre-commit config also updated to run unit and validation tasks separately and apply coverage only to unit tests.

furqan463:
> pre-commit config also updated to run unit and validation tasks separately

https://github.com/PowerGridModel/power-grid-model-io/pull/451#issuecomment-4993097917 @mgovers PS: now I can remeber, the discussion was regarding pre-commit not running validation tests. :relieved: :smile:

Jerry-Jinfeng-Guo (approved):

PR #492, opened 2026-09-06
fix: batch Vision GUID number columns

be-student:
Fixes #468.

Builds all generated Number columns in one concat operation instead of repeatedly inserting them. This removes pandas fragmentation warnings and preserves each GUID/Number pair's order, including MultiIndex inputs and existing Number columns.

Validation: 1,444 tests pass; focused tests, Ruff, MyPy, REUSE, and full pre-commit pass.

PR #500, opened 2026-09-19
Batch Vision GUID columns without fragmentation warnings

### LINKED ISSUE #468: [IMPROVEMENT] fix warnings around performance issues
### Problem faced

We use the converter for vision file exports. 

The logging shows a performance warning:
```
/.venv/lib/python3.13/site-packages/power_grid_model_io/data_stores/excel_file_store.py:249: PerformanceWarning: DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
  data.insert(guid_column_pos + 1, new_column_name, data[guid_column].apply(self._uuid_cvtr.query))
```

### Improvement proposal

See if we can rewrite the logic in such a way that the warning doesn't get triggered anymore. 
Speed wise the converter isn't quick, but performance is accaptable for us. So no rush to improve speed on our side, but would be nice to have less warnings/errors either way. 

### Community Meeting

- [ ] I wish to discuss this improvement in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

appleweiping:
Fixes #468. Reconstructs the Excel columns once instead of repeatedly calling DataFrame.insert. Generated Number columns keep source order; existing Number columns are replaced in place. The regression tests cover 105 GUID columns, two-level headers, special-sheet Subnumber mapping, and non-string labels.

This is a fresh branch from current main after #492 closed, with one DCO-signed commit verified by GitHub.

Validation: 1,447 tests passed on Windows with Python 3.14.5; coverage 98.61%. Mypy passed for 93 source files, and Ruff check and format passed. I did not run a real Vision workbook throughput benchmark.

AI assistance: OpenAI Codex was used for implementation and verification.


Jerry-Jinfeng-Guo:
Hi @appleweiping , based on your response in #468 (that you may be slow), should we start reviewing this PR? Asking because this PR is now in review, not a draft.

appleweiping:
Hi Jerry, yes, please feel free to start reviewing it. The
implementation and tests are ready for review now. Thanks!

tests/unit/data_stores/test_excel_file_store.py
```diff
@@ -302,6 +303,69 @@ def test_remove_unnamed_column_placeholders__empty():
...
+    store = ExcelFileStore()
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", pd.errors.PerformanceWarning)
+        result = store._process_uuid_columns(data=data, sheet_name="Other")
+
+    assert list(result.columns) == [name for i in range(105) for name in (f"Field{i}GUID", f"Field{i}Number")]
+    assert result.loc[0, "Field0Number"] == 0
+    assert result.loc[0, "Field104Number"] == 104
+
```
Jerry-Jinfeng-Guo, line 317: Could you add an assertion here also for GUID's? e.g.:

```suggestion
assert result.loc[0, "Field104GUID"] == 'id-104'
```

tests/unit/data_stores/test_excel_file_store.py
Jerry-Jinfeng-Guo, line 1: The tests in this file not only tests the pandas fragmentation warning at larger than 100 load but also tests a more reasonable ordering of columns, which was largely overlooked in the old implementation. To make this test more comprehensive, I would suggest add tests for the following:
 - Empty DataFrames with GUID columns.
 - `None`/`NaN` GUID values
 - Duplicate GUID or derived-number labels
 - Multiple calls across sheets using the same `UUID2IntCvtr`
 - Explicit input-mutation behavior vs old modificatyion in place approach

Jerry-Jinfeng-Guo (commented):
I would expect the peak RAM footprint of this reconstruction approach to be higher than the old mutation in place approach. Can you verify whether this might be an issue?

appleweiping:
Hi Jerry,

Thanks for the careful review. I agree that the memory overhead and
the additional test cases need to be addressed before this is ready to
merge.

I’m not able to work through the implementation revisions promptly at
the moment. Would you or another maintainer be open to picking up the
remaining changes on this PR? I would really appreciate the help, and
I’m happy for you to build on the existing work and adjust the
approach.

Here are the technical findings and suggestions that may help with that.

On memory, a preliminary isolated comparison of the helper logic
showed a median sampled peak RSS of about 389 MiB for the current
reconstruction versus 182.5 MiB for the old implementation. This used
50,000 rows, 105 GUID columns and 20 numeric columns, with three runs
per implementation on Linux using pandas 2.2.3 with Copy-on-Write
disabled. This was a synthetic stress case, not a complete Vision
workbook conversion, so it demonstrates a potential regression rather
than its impact on typical workloads.

Simply changing `pd.concat` to `copy=False` does not look sufficient:
in the isolated checks, it reduced memory usage but allowed changes to
the result to affect the input, and a subsequent column insertion
could trigger the fragmentation warning again.

For the implementation, I would first make the output column positions
and replacement rules explicit before constructing the result. The
current `next(...)` lookup only replaces the first matching Number
column, and the additions logic does not account for different GUID
labels generating the same target name. Those cases need to be handled
consistently with the duplicate column preprocessing in `load()`,
rather than silently keeping stale values or introducing duplicate
targets.

For the tests, I agree with all the cases you listed. I would compare
the original GUID columns in full, check that empty inputs still
receive the expected derived columns, and verify that repeated calls
preserve the shared converter’s mappings. I would also keep the
existing distinction between `None` and `NaN` in `UUID2IntCvtr` for
this PR, rather than changing identifier semantics as part of the
warning fix. The mutation tests should cover both changes during the
call and whether later edits to the result affect the input.

The loader already uses the returned DataFrame, so returning a new
object seems compatible with that call site. The remaining question is
how to construct it with acceptable peak memory while preserving the
intended behavior. Any replacement would still need validation through
actual sheet loading and the Vision conversion tests.

I haven’t validated a final replacement implementation, so these are
notes for the next revision rather than a completed fix. I understand
if taking this on needs to wait for maintainer availability.

Thanks again for your help.

Best,
Weiping

Jerry-Jinfeng-Guo:
Thanks @appleweiping ! We will discuss this internally and le you know how we intend to follow up!

Jerry-Jinfeng-Guo:
Hi @appleweiping , we would like to update you with the following:
 - Our internal tests show RAM footprint is fine with data considered large on our side
 - We are fine with waiting for you to finalize the PR

That being said, we think it's needed to add a stress test with synthetic data (almost unreasonably huge) to show the RAM footprint is indeed manageable. 

We look forward to your input!

